PR #16, opened 2017-10-24
make sure dependent targets get the include directory added correctly

arvidn:
This fix makes dependent targets automatically get the cppkafka include directory added when declaring cppkafka as a dependency.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,6 +27,7 @@ add_library(cppkafka ${CPPKAFKA_LIBRARY_TYPE} ${SOURCES})
 set_target_properties(cppkafka PROPERTIES VERSION ${CPPKAFKA_VERSION}
                                           SOVERSION ${CPPKAFKA_VERSION})
 target_link_libraries(cppkafka ${RDKAFKA_LIBRARY})
+target_include_directories(cppkafka PUBLIC ${PROJECT_SOURCE_DIR}/include)
 
 install( 
     TARGETS cppkafka


src/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ add_library(cppkafka ${CPPKAFKA_LIBRARY_TYPE} ${SOURCES})
 set_target_properties(cppkafka PROPERTIES VERSION ${CPPKAFKA_VERSION}
                                           SOVERSION ${CPPKAFKA_VERSION})
 target_link_libraries(cppkafka ${RDKAFKA_LIBRARY})
+target_include_directories(cppkafka PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../include)
```
mfontanini: I think it's better to use `${PROJECT_SOURCE_DIR}` here so you don't have to traverse the path back.

arvidn: heh, I'm a cmake noob. I was just copy-pasting the include path used internally (a few lines above). Your suggestion seems like a good idea

mfontanini: Yeah I actually want to go over these CMake files and clean them up soon, I know a little bit more about CMake now than I did when I wrote those.

arvidn: actually, the problem with that is that I'm pulling in cppkafka as a submodule and building it as part of my project tree. ``${PROJECT_SOURCE_DIR}`` will then point to *my* root, not cppkafka's root.

arvidn: I consider this to be a very useful (if not the most useful) use case for linking against other cmake libraries

mfontanini: Are you using the project root when you add as a submodule? According to [this](https://cmake.org/cmake/help/v3.0/variable/PROJECT_SOURCE_DIR.html) it should point to the directory where you last called `project` which is in cppkafka's root CMake file. If this doesn't work, no problem. I'll clean this up soon anyway so I'll see if there's an alternative.

arvidn: hm. I may have made some incorrect assumptions. I'll take a closer look

arvidn: I had a typo when testing it. I force-pushed the fix into this PR

arvidn:
while on the topic of cmake files. another issue I'm having that I don't know how to solve is the ``find_package(RdKafka)``. I have librdkafka as a submodule built with cmake as well, but I can't figure out how to make it available for ``find_package``.

mfontanini:
You should define `RDKAFKA_ROOT` before importing cppkafka. I'm also using cppkafka as a submodule on a bunch of projects and that seems to work. e.g.

```cmake
set(RDKAFKA_ROOT something/third_party/librdkafka)
add_subdirectory(something/third_party/cppkafka)
```

I'm actually using `ExternalProject_Add` rather than `add_subdirectory` but I think that should work.

arvidn:
I take it you meant ``RDKAFKA_ROOT_DIR``. either way, the ``findRdKafka.cmake`` file insists that I define both ``RDKAFKA_INCLUDE_DIR`` and ``RDKAFKA_LIBRARY``. Part of the problem (I think) is that in the librdkafka project, there is no include directory called ``librdkafka`` containing all the headers, where it seems cppkafka expects this, not just in the code but also the cmake test program that checks that it can find the library.

PR #18, opened 2017-10-25
fix typo in consumer documentation

arvidn:
that's a std::vector


mfontanini:
I don't know how I came up with `count`. Thanks!

PR #21, opened 2017-11-21
Check for existing custom target: uninstall

spektom:
When including `cppkafka` in a CMake project containing other project defining the same target (`curl`, specifically, which defines the same target [here](https://github.com/whoshuu/curl/blob/3ea76790571c1f7cf1bed34fabffd3cc20ad3dd3/CMakeLists.txt#L1293)) I get the foollowing error:

```
add_custom_target cannot create target "uninstall" because another target  with the same name already exists
```

This if condition fixes the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,13 +86,15 @@ else()
     message(STATUS "Disabling tests")
 endif()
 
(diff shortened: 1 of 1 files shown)


mfontanini:
Thanks!

PR #27, opened 2017-11-29
Update README.md

alkenet:
Should there not be a flush at the end? In my case, the application quit without the message ending up on Kafka without the flush.


mfontanini:
Yep, that's true. Thanks!

PR #39, opened 2018-02-08
destruct callbacks (and their closures)

arvidn:
to ensure there are no reference cycles to rd_kafka objects when destructing the consumer. This solves a potential hang in rd_kafka_destroy().
Here's a related ticket leading me to this patch: https://github.com/edenhill/librdkafka/issues/1117


mfontanini:
Thanks!

PR #40, opened 2018-02-08
improve performance of Consumer::poll_batch()

arvidn:
by allocating the results array once, rather than incremental growth.
Also avoid checking every single pointer in the returned array, just use the returned counter instead.


mfontanini:
Thanks!

PR #49, opened 2018-04-07
Implement configuration type safety for librdkafka settings

accelerated:
Note that this change relies on the following librdkakfa [pull request](https://github.com/edenhill/librdkafka/pull/1756)

### Motivation:

Prior to this change it was impossible to determine which settings were necessary or would apply to a certain type. For example one could very well set a Producer setting on a Consumer object and if the setting name & type were correct, the library would accept it, although it would make little sense and would not take effect. This put an additional burden on the user to verify the documentation each time. 

### Description of changes:

One PR was made to librdkafka in order to provide an API to get configuration metadata at runtime. This allows any higher level library to retrieve this list and apply appropriate checks which is exactly what the current change does. In order to maintain full backwards compatibility with previous releases of the code, certain design decisions were carefully considered for preserving the type names and signatures. Had this requirement not been important, the overall changes could have been somewhat simpler and more elegant.

The main changes are as follows: 

* Some type traits (see `cppkafka/traits.h`) were introduced to differentiate between various consumer/producer behaviors when consuming settings. Traits also allow for future customization so it's a nice to have feature. 
* If the wrong setting or callback is applied to the inapropriate object type, then the code will throw. 
* The file `cppkafka/types.h` contains the new types (as well as the legacy types). Note that new type naming can always be changed if deemed necessary, I tried to keep things as consistent as possible.
* Due to some templatization, an `impl` directory has now been created both under _include_ and _utils_ directories to contain the 
implementation of all class methods. This helps keep the class definition header simple and clean. Some classes were also renamed, however the legacy types have been aliased back to the same names. As such any old code should still compile as-is.
* Single include header was created i.e. `cppkafka/cppkafka.h`. This makes using the library much easier. 
* Examples and tests were left with the legacy types, and they should probably be changed eventually to reflect usage of the new types. Legacy types have no runtime checking and as such behave as previously.
* In order to validate settings at runtime a cache object queries librdkafka and stores all the metadata. This cache is then searched every time a new setting is applied. (see `cppkafka/configuration_cache.h`)
* Keeping producer and consumer settings separate, the configuration and topic configuration classes were enhaced with traits which determine their scope at runtime. As such a Producer can only use producer handle configuration and producer topic configuration settings. Similarly for the Consumer object.

# Scope of settings

A producer configuration object can use:
* Any global settings
* Producer specific settings and callbacks

A consumer configuration object can use:
* Any global settings
* Consumer specific settings and callbacks
* Consumer group settings

A producer topic configuration object can use:
* Any global topic settings
* Producer topic specific settings

A consumer topic configuration object can use:
* Any global topic settings
* Consumer topic specific settings

A producer object can use:
* Producer configuration objects
* Producer topic configuration objects

A consumer object can use:
* Consumer configuration objects
* Consumer topic configuration objects

**Note: As per above, topic-specific configurations cannot be applied to global configuration objects and vice versa.**

Although the overall changes seem a lot, there is very little functionality change. However the class structure is a little bit better defined now and allows for more flexibility in the future as more traits could be applied and used.


mfontanini:
Thanks, I appreciate the amount of effort you put into this PR. However, instead of basically changing the entire library just to check whether a configuration option is valid for a specific rdkafka handle type, wouldn't it be better to just check change the current `Consumer`/`Producer` class to check that the given `Configuration` object used on their constructors contains attributes that fit the handle being instantiated? Same for `TopicConfiguration` when setting attributes. The only advantage I can see in your change is you would know that the attribute doesn't make sense earlier, that is when setting it on the config rather at the handle instantiation time, but the outcome should be the same: the user knows he messed up.

This is an insane amount of code modified to just introduce a change that's pretty much invisible for the user, while at the same time making the code harder to follow and breaking the ABI while doing so.

At the same time I'm wondering whether `rdkafka` should solve this instead. Shouldn't it be checking whether the options passed in make sense instead of this (and possibly other) wrappers?

accelerated:
Hi, yes of course this was my first thought, however in order to maintain backwards compatibility it's very hard to do just that, and unfortunately the rdkafka code itself does not easily lend itself to do that. I will give you some examples why you can't check at runtime as you suggested:

* For instance the Consumer/Producer objects take in the constructor a Configuration object. If this was a list of ConfigurationOptions, which you would then set in the librdkafka library, that would be easy to check, instead the Configuration object itself has already set all the settings into the library before you even pass it to the constructor. How then can you verify? Recall the entire settings of librdkafka and check what's set and what's not and then validate? So already there it's not too easy.

* Same thing applies for the TopicConfiguration class and the default topic configuration. All settings are done beforehand and then passed into the handles. 

* Configuration class contains callback apis which apply for both Producers and Consumers. Again this setting is done prior to handle construction. Not only that but how can you enable/disable the wrong api (say offset committer callback on a config that's intended for a producer) before knowing what the configuration object will be used for?

* Enforcing this in librdkafka, is also not desirable as it's a C api and the setters and getters for options are very class agnostic. You would have to essentially pass the intended scope of the setting as a parameter to the API in order to validate, which pretty much indicates that you already know what you're doing beforehand. I did look at the librdkafka code quite a bit and I don't think it's that easy. 

* The only way IMHO to cleanly do this is to separate intent at the class level. Then, any user of your library can easily know what classes to use and how to combine them together without thinking too much or shooting themselves in the foot. Also turning off certain callback setter/getter methods in the implementation classes at compile time really gives you a clean interface. I don't think the ABI has changed, why do you think it has?

* The code changes seem a lot, but actually it's very little and they are very localized. Some stuff was moved around but that's about it. Maintaining 100% backward compatibility (which is what i did) is not easy. As I mentioned earlier, a simpler design change which would break the previous version would be to modify the existing classes directly and keep Consumer/Producer as-is. Some other solutions could be overloading most methods which deal with configuration and have different argument types. That's even uglier IMHO.

accelerated:
PS: Regarding ABI changes, yes some interfaces and exports are now different, but I think it's a small price to pay and the majority of people will recompile anyway if they pull the latest version of your code. 

On the other hand i seem to have a newer compiler and i can't seem to figure out why the librdkafka enums are not recognized in the regression builds...? Ideas?

accelerated:
I would also like to address this concern:

>This is an insane amount of code modified to just introduce a change that's pretty much invisible for the user, while at the same time making the code harder to follow and breaking the ABI while doing so.

The purpose of any library, much so a wrapper one, is to abstract as much functionality for the end user as you can and simplify usage. All logic is thus _invisible_ and the more failsafes and guards you provide the more value-add you offer. If on the other hand you write a library where you place half the burden on the user, then I feel the library becomes a liability for mistakes and hard to debug production issues (e.g. some setting you provided which was silently swallowed and never took effect). So this is a bit of a moot point.

Although I agree that the code is "harder" to read, I feel this is more of a documentation issue. STL and Boost are pretty opaque to read and they are still the most used C++ libraries out there. It's all about proper examples and documentatiin, which you already provide. Also you have written some pretty advanced metaprogramming stuff (aka hard to read for a novice) in the utils classes and that does not prevent people from using them. 

If librdkafka accepts my PR, then you can feel free to reject this PR and implement it any which way you choose. I've written throw away code before, but in this case I feel it's a shame as you get lots of benefits from metaprogramming (and I think you can appreciate these powerful features as you seem quite well versed).

mfontanini:
Addressing all of your raise points individually:

> For instance the Consumer/Producer objects take in the constructor a Configuration object. If this was a list of ConfigurationOptions, which you would then set in the librdkafka library, that would be easy to check, instead the Configuration object itself has already set all the settings into the library before you even pass it to the constructor. How then can you verify? Recall the entire settings of librdkafka and check what's set and what's not and then validate? So already there it's not too easy.

`ConfigurationBase` could keep track of all option names (just the names) that were set by the user and then you would just check those instead of the entire list of attributes that rdkafka returns for every configuration handle. When a `Consumer`/`Producer` is created, a check can be done to make sure all user set options in the config make sense.

> Same thing applies for the TopicConfiguration class and the default topic configuration. All settings are done beforehand and then passed into the handles.

Similar solution as above.

> Configuration class contains callback apis which apply for both Producers and Consumers. Again this setting is done prior to handle construction. Not only that but how can you enable/disable the wrong api (say offset committer callback on a config that's intended for a producer) before knowing what the configuration object will be used for?

Same solution as above, attributes which map to callbacks still have a string name in rdkafka so you can just mark this option as "user set" and then check it the same way. Sure, you won't get a compilation error but you will still get an exception when instantiating the consumer/producer so this won't go unnoticed.

> Enforcing this in librdkafka, is also not desirable as it's a C api and the setters and getters for options are very class agnostic. You would have to essentially pass the intended scope of the setting as a parameter to the API in order to validate, which pretty much indicates that you already know what you're doing beforehand. I did look at the librdkafka code quite a bit and I don't think it's that easy.

I don't understand this. The solution I'm proposing could be implemented in rdkafka in a similar way: when creating the consumer/producer handles, you could check if the options make sense. Not sure if rdkafka keeps track of modified options but that would be a simple change.

> The only way IMHO to cleanly do this is to separate intent at the class level. Then, any user of your library can easily know what classes to use and how to combine them together without thinking too much or shooting themselves in the foot. Also turning off certain callback setter/getter methods in the implementation classes at compile time really gives you a clean interface.

For any attributes except callback ones, what really is the advantage besides the fact that you know you messed up when creating the configuration handle rather than the consumer/producer itself? How is doing this:

```c++
Configuration config = {
    { "consumer.specific.config", "foo" }
};
Consumer consumer(config); // oops, throws
```

More error prone than this?

```c++
KafkaConsumerConfig config = {
    { "consumer.specific.config", "foo" }
}; // oops, throws
```

The result is the same: an exception is thrown on runtime.

> The code changes seem a lot, but actually it's very little and they are very localized. Some stuff was moved around but that's about it. Maintaining 100% backward compatibility (which is what i did) is not easy. Your library is very small, with few classes only and most of the logic is actually providing settings and callbacks. So any change in that department would seem a lot.

I'm not against changing things. I just don't think it's justifiable to make all of these changes for something that can be fixed in a much simpler way. 

> Also conceptually it's a lot easier to think in terms of Producer/ProducerConfig/ProducerTopicConfig & Consumer/ConsumerConfig/ConsumerTopicConfig for someone who does not know the code well.

It may just be me but I don't see how from a user perspective this actually simplifies anything. The sole purpose of this change is to add error checking on runtime. Yes, a user can't use a `ProducerConfig` where a `ConsumerConfig` should be used (compile time error) but they can still create a `ConsumerConfig` and set a producer config on it, which ends up failig on runtime. The new config classes are just there to support the changes you want to do and not because they're really making things better. If setting a wrong attribute failed on compile time (besides callback ones), then that would be another thing but that's not the case (and it's not possible). Moreover, now there's 2 consumer classes, 2 producer classes, 3 config classes with the word "topic" on them, etc. This is going to be way more confusing for a user than the current situation.

> The purpose of any library, much so a wrapper one, is to abstract as much functionality for the end user as you can and simplify usage. All logic is thus invisible and the more failsafes and guards you provide the more value-add you offer. If on the other hand you write a library where you place half the burden on the user, then I feel the library becomes a liability for mistakes and hard to debug production issues (e.g. some setting you provided which was silently swallowed and never took effect). So this is a bit of a moot point.

The solution I proposed still would fail on runtime so I don't think this argument is valid. As it currently stands, yes, you can set options that will be silently discarded by rdkafka (and again, I'm wondering if this shouldn't be fixed at the rdkafka level rather than on wrappers). Also, if this is really something that can cause issues when debugging production issues then you should point that out on the rdkafka PR and propose a way to handle this over there as pretty much any project that uses rdkafka either explicitly or implicitly through the various wrappers that exist is likely affected by it.

accelerated:
In reference to rdkafka implementing these checks, I would have to look at the code and see. Last time I checked changing anything in there is close to performing heart surgery, not exactly the easiest to grasp code. Also the author is not super responsive either. There's lots of open issues and not sure how maintainable it is, so i would rather keep this in application space. If you could nudge the author to accept the PR i submitted we'd at least be half way there. 

Back to the curent proposal. I'll just make a correction that there's _conceptually_ a single Producer and Consumer class, not two. The other one is clearly marked as legacy, and I don't think anyone will create a Consumer class with Producer traits. All this is well documented in the `types.h` file via aliases. In the future, should you want to implement kafka streams or kafka connect features, it will be much simpler if you already have traits. Also the generic type can be eventually removed. Something to think about...

Regarding your proposal of recording options in the configuration classes, it's definitely feasible but it's being _reactive_ vs being _proactive_. It's the same as building a boat and then realizing your client wants a car and throwing away all that work, instead of checking from the get-go. Also, if you have a configuration class and 100 consumers, you will be performing the validation 100 times. Ok maybe you can keep state if the checks pass once, but then you have one state per consumer/producer type (imagine all settings are global for example so they could apply to both) and another two states if it's a topic config or not (remember this is a base class) so now the logic starts to get more complicated as you have 4 states. 
In my case, besides the options cache which will have to be used in this new proposal as well, my only functional changes are 3 lines of code in the `set()` method of the `Configuration` and `TopicConfiguration` classes. And maybe some cosmetic polymorphic changes in the `ConfigurationBase` class. That's it! The rest is just template stuff and search/replace text patterns. I think that besides the "omg there are too many lines changed!" which might be intimidating at first, i really haven't heard any real technical argument against it besides being a large scope and the ABI. Sure there's more than one way to skin the cat and in the end it's your decision to make, ultimately.

Ok bottom line is i can re-implement the options recorder functionality with state if this is your final preference. But i will not proceed until the librdkafka PR is approved as i don't want to work twice, so if you can help in that department that would be great!!

mfontanini:
Regarding librdkakfa, the author seems to reply on a daily basis, there's just too many tickets. I'm subscribed to the github updates and the amount of dumb tickets that get created there every day is ridiculous. Most of them from people who just don't know C and try to use rdkafka, which obviously won't go well.

Going back to the proposal, I know it's just a single class template instantiated with different template arguments, but from a user's perspective there's now 2 types of consumers/producers.

Checking if the options on a config are valid involves a single lookup and a couple of comparison per option set by teh user. Even if you have 100 consumers and say... 10 options per consumer? 20? That's 2k lookups in a map which is completely negligible. There's no need for a cache at all.

Overall I'm really trying to understand why you want to fix this so bad. Have you had issues with this by setting an attribute that was ignored? I don't imagine this is a very common thing. If you're trying to tweak rdkafka down to some very specific parameters, you probably want to know what you're doing. It's not like you'll go "oh let me tweak the coordinator query interval" but you don't know that really does.

Finally, I've written about a dozen applications using cppkafka at my current workplace. I actually wrote cppkafka because both rdkafka and rdkafka++ were a pain to work with and I knew I'd be using kafka a lot. In some apps I actually have some base configuration that I use as a template for all other configs (both producer and consumer). This base configuration has a mixed of global, consumer and producer only configs so in my own code I go against this proposal. Why do I do that? Why not? If I have 1 or 2 producers, I don't want to have to set producer specific options on a separate configuration. This way I know the common, shared, base configuration is defined on a single spot, regardless of whether it's going to be used for consumers or producers, so if I have a doubt or I want to add some new option and make sure it's used everywhere, I just need to add it on a single place regardless of what its scope is.

accelerated:
Hi, I have made all [changes](https://github.com/edenhill/librdkafka/pull/1760) inside librdkafka. The author preferred this way and I also removed getting the list of options at runtime as per his request since he didn't want to expose it to the outside world. If you could quickly code review them and provide feedback that would be great. I will have to apply these changes to your code, mainly checking return values from various librdkafka api. So changes in cppkafka will be minimal.

accelerated:
@mfontanini Hi, so based on the latest [PR for librdkafka](https://github.com/edenhill/librdkafka/pull/1764), the default scope check behavior is disabled. It can be turned on _globally_ via `rd_kafka_enable_conf_scope_validation()`. Since this is a global setting, where do you think it should go? As a static method part of the Configuration class? But this will also affect TopicConfiguration. I could put it in the ConfigurationBase but that's templetized. Alternatively I could have a standalone function somewhere. Any suggestions?

mfontanini:
Whatever change you end up applying here won't be merged into cppkafka until rdkafka actually publishes a release that includes your changes as the code here needs to conditionally compile to something if the changes are present in rdkafka. I don't think there's a point in rushing this, even more so when your rdkafka PR hasn't even been reviewed (I know he's reviewing it now but still).

accelerated:
Yes agreed.

PR #51, opened 2018-04-19
Added cmake option for disabling tests

EvilBeaver:
It's useful to have an option for disabling tests when building on CI server.

# Use Case:

* Imagine you have a project which uses cppkafka as a submodule
* You build project with CI server and for full-source build you do a git checkout with a recursive submodule update
* Now after checkout you have googletest submodule also cheсked out and cppkafla will be doing a tests build
* But as a submodule, cppkafka is always checked out on one specified commit and this means it's code is unchanged and stable. So it doesn't require any continuous testing.
* Having cppkafka tests built on your daily builds all you have is a time waste

This PR introduces CMake option CPPKAFKA_BUILD_TESTS which is enabled by default and old behavior is untouched. But it's easy to disable building tests with this option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib)
 
 # Build output checks
 option(CPPKAFKA_BUILD_SHARED "Build cppkafka as a shared library." ON)
-option(CPPKAFKA_BUILD_TESTS "Build cppkafka tests." ON)
+option(CPPKAKFA_DISABLE_TESTS "Build cppkafka tests." OFF)
 if(CPPKAFKA_BUILD_SHARED)
     message(STATUS "Build will generate a shared library. "
             "Use CPPKAFKA_BUILD_SHARED=0 to perform a static build")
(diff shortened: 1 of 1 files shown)


mfontanini:
Thanks for this PR! The tests are not included in the `all` Makefile target so I don't think normally this would be an issue since the build will just skip them.

Anyway, this is a reasonable change and I can see its usefulness. The only thing I'd change is the option name. I normally prefer names in the form of `<project>_DISABLE_<feature>` for these types of parameters. Maybe `CPPKAKFA_DISABLE_TESTS`? Otherwise this looks like `CPPKAFKA_BUILD_SHARED` but has a very different meaning: the latter controls _how_ something is built and the former controls _what_ is being built.

On a similar note, there should probably be a similar flag to disable examples as well, but that can be tacked later.

EvilBeaver:
Yes, i chose parameter name looking at CPPKAFKA_BUILD_SHARED option. 
I build it using cmake and Visual Studio so Makefile and it's targets are not involved. Okay, I'll change option name

EvilBeaver:
done

mfontanini:
Fair enough, I didn't consider that. Looks like I typo-ed the option name and you copy pasted it. It says `CPPKAKFA`. Could you correct it? Sorry about that!

EvilBeaver:
Copypastes are evil :) Fixed

mfontanini:
lol yeah, also I should have probably worn my glasses while I replied in the morning. Thanks, I'll merge this later!

PR #52, opened 2018-04-19
Added commit for current assignment and message state checks

accelerated:
### Description

* Added `commit()` and `async_commit()` methods in the consumer class for committing on the current partition assignment. This is supported directly by librdkafka api. 
* Added null handle checks in the `Message` class.
* Changed `std::vector<TopicPartition>` -> `TopicPartitionList` alias for consistency in the code.
* Added Consumer object reference in partition assignment callbacks. This allows access to the object being re-assigned and is more consistent with all other callbacks which pass-in Producer or KafkaHandleBase as first parameter.

### DIFF
--- a/src/consumer.cpp
+++ b/src/consumer.cpp
@@ -116,6 +116,14 @@ void Consumer::unassign() {
     check_error(error);
 }
 
+void Consumer::commit() {
+    commit(nullptr, false);
+}
+
+void Consumer::async_commit() {
+    commit(nullptr, true);
+}
+
 void Consumer::commit(const Message& msg) {
     commit(msg, false);
 }
@@ -125,11 +133,11 @@ void Consumer::async_commit(const Message& msg) {
 }
 
 void Consumer::commit(const TopicPartitionList& topic_partitions) {
-    commit(topic_partitions, false);
+    commit(&topic_partitions, false);
 }
 
 void Consumer::async_commit(const TopicPartitionList& topic_partitions) {
-    commit(topic_partitions, true);
+    commit(&topic_partitions, true);
 }
 
 KafkaHandleBase::OffsetTuple Consumer::get_offsets(const TopicPartition& topic_partition) const {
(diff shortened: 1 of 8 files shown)


src/consumer.cpp
```diff
@@ -238,34 +270,43 @@ void Consumer::close() {
...
 }
 
-void Consumer::commit(const TopicPartitionList& topic_partitions, bool async) {
-    TopicPartitionsListPtr topic_list_handle = convert(topic_partitions);
+void Consumer::commit(const TopicPartitionList* topic_partitions, bool async) {
     rd_kafka_resp_err_t error;
-    error = rd_kafka_commit(get_handle(), topic_list_handle.get(), async ? 1 : 0);
+    error = rd_kafka_commit(get_handle(),
+                            !topic_partitions ? nullptr : convert(*topic_partitions).get(),
+                            async);
```
mfontanini: Can you keep the explicit conversion to 1/0 here? I prefer to know that's what rdkafka expects, even if `bool`s technically can only have those values.

accelerated:
@mfontanini thanks for your quick review.

Ok with all your comments. Will split the PRs into two (callback enhancement separate). However I do urge to reconsider your stand on the callback (maybe ask others what they think?). I know you can theoretically bind every single consumer instance and register that callback with itself, but that adds another layer of complexity which is unnecessary and now I have to call bind every single time which is awkward (some programmers may even devise more complex solutions to solve this). Since the callback is raised in the context of the object itself, it's only natural to get a reference to the context in which it happens. Not only it gives programmers the chance to **do** something with the instance (we definitely have a use for it hence my change), but it also makes things _consistent_ with other callbacks (even if they are in the context of a configuration object). I think your library does a fantastic job at abstracting _librdkafka_ and as such it should continue to making programmers' lives easier. I know it makes the Consumer class api a bit ugly (2 callback registrations, 2 getters, etc) but we could put a compile warning for deprecation and eventually retire it.

PR #54, opened 2018-04-20
Added Consumer object reference to partition callbacks

accelerated:
Added contextual reference to the consumer instance in the partition assignment callbacks. See [here](https://github.com/mfontanini/cppkafka/pull/52#issuecomment-383114054) for reasoning. Hoping this one will be accepted.

### DIFF
--- a/include/cppkafka/consumer.h
+++ b/include/cppkafka/consumer.h
@@ -98,8 +98,11 @@ class TopicConfiguration;
 class CPPKAFKA_API Consumer : public KafkaHandleBase {
 public:
     using AssignmentCallback = std::function<void(TopicPartitionList&)>;
+    using PartitionAssignmentCallback = std::function<void(Consumer&, TopicPartitionList&)>;
     using RevocationCallback = std::function<void(const TopicPartitionList&)>;
+    using PartitionRevocationCallback = std::function<void(Consumer&, const TopicPartitionList&)>;
     using RebalanceErrorCallback = std::function<void(Error)>;
+    using PartitionRebalanceErrorCallback = std::function<void(Consumer&, Error)>;
 
     /**
      * \brief Creates an instance of a consumer.
@@ -134,7 +137,12 @@ public:
      *
      * \param callback The topic/partition assignment callback
      */
-    void set_assignment_callback(AssignmentCallback callback);
+    void set_partition_assignment_callback(PartitionAssignmentCallback callback);
+    
+    /**
+     * @remark Deprecated. Use set_partition_assignment_callback()
+     */
+    DEPRECATED void set_assignment_callback(AssignmentCallback callback);
 
     /**
      * \brief Sets the topic/partition revocation callback
@@ -149,7 +157,12 @@ public:
      *
      * \param callback The topic/partition revocation callback
      */
-    void set_revocation_callback(RevocationCallback callback);
+    void set_partition_revocation_callback(PartitionRevocationCallback callback);
+    
+    /**
+     * @remark Deprecated. Use set_partition_revocation_callback()
+     */
+    DEPRECATED void set_revocation_callback(RevocationCallback callback);
 
     /**
      * \brief Sets the rebalance error callback
@@ -160,7 +173,12 @@ public:
      *
      * \param callback The rebalance error callback
      */
-    void set_rebalance_error_callback(RebalanceErrorCallback callback);
+    void set_partition_rebalance_error_callback(PartitionRebalanceErrorCallback callback);
+    
+    /**
+     * @remark Deprecated. Use set_partition_rebalance_error_callback().
+     */
+    DEPRECATED void set_rebalance_error_callback(RebalanceErrorCallback callback);
 
     /**
      * \brief Subscribes to the given list of topics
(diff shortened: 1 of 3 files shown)


accelerated:
I think your particular implementation (with the consumer as a member) is fine, except that it may not be what other people are doing. This is a general purpose library, so use-case will vary as it is in my case i.e. no wrappers. 
It's not the `ref` issue with bind (that's just a limitation of using bind), it's just the extra work to be done when in reality you have all the info there and you can just pass it as a parameter. Having an extra parameter that you don't use is hardly an issue compared to binding each instance. In the same token I may not be using the `Producer` or `KafkaHandleBase` references pass-in by the other callbacks. In the deprecated variant, there is extra work to be done while in the proposed API none. That's where i'm trying to get at. It's a small limitation of the library which could be easily addressed. As usual, your call. :)

mfontanini:
> it's just the extra work to be done when in reality you have all the info there and you can just pass it as a parameter

One could say the exact same thing about your solution. Why bind the consumer you're setting the callback on? You have the consumer right there when setting your callback so you could just bind it.

I'm thinking of asynchronous event libraries I've used and can only remember boost.asio right now but it follows this same idea: no callbacks are called with a reference to the object that triggers the event. Why would the class that generates an event (`Consumer` in this case) need to pass itself to a callback that was explicitly handed to it? I think it makes more sense to not do that and if the user actually needs that, they are free to do so.

accelerated:
> I'm thinking of asynchronous event libraries I've used and can only remember boost.asio right now but it follows this same idea: no callbacks are called with a reference to the object that triggers the event. Why would the class that generates an event (Consumer in this case) need to pass itself to a callback that was explicitly handed to it? I think it makes more sense to not do that and if the user actually needs that, they are free to do so.

I agree somewhat, but you have to realize that inside the callback, you have no other way of knowing what context you're executing in. The callback is not a member function in which normally you can use `this` and be done with it. Here the callback is anything the user passes in. My code right now looks like this:
```
//bind partition callbacks
cppkafka::Consumer consumer(consumerConfig);
Callbacks::AssignmentCallback assign = std::bind(Callbacks::Assignment, std::ref(consumer), _1);
Callbacks::RevocationCallback revoke = std::bind(Callbacks::Revocation, std::ref(consumer), _1);
Callbacks::RebalanceErrorCallback rebalance = std::bind(Callbacks::RebalanceError, std::ref(consumer), _1);
consumer.set_assignment_callback(assign);
consumer.set_revocation_callback(revoke);
consumer.set_rebalance_error_callback(rebalance);
```
and I would prefer it to look like:
```
cppkafka::Consumer consumer(consumerConfig);
consumer.set_partition_assignment_callback(Callbacks::Assignment);
consumer.set_partition_revocation_callback(Callbacks::Revocation);
consumer.set_partition_rebalance_error_callback(Callbacks::RebalanceError);
```
Someone looking at the code would wonder why you need to bind the object and then register the function unto itself? Plus it feels _klunky_. If you are on the fence about this, perhaps you can ask other users in your company what they think and we leave it to a consensus decision. I'm up for democracy :)

accelerated:
I guess I could pass-in lambdas and capture the consumer by ref, but none of my other callbacks are lambdas and I'm even more reluctant to use this approach.

mfontanini:
I don't see how this is any different from what other libraries like boost.asio have, and I don't see them doing this. If _you_ want to get the consumer in the callback, you're free to do so by binding it/capturing it. If you just couldn't do that for some reason, then this would be a different situation. You can't try to deprecate a part of the API because you don't want to type `ref(consumer)` 3 times in your code.

accelerated:
Well, i'm not sure about what boost::asio does (I used it once only) or other libraries, it may well be that their callback parameters are self-sufficient? In this case there's definitely a use. `ref (consumer)` has nothing to do with this, its about `bind` and it's not 3 times it's 3 times _per consumer_. We don't have to deprecate the other functions, it was suggestive as that's more or less standard procedure when evolving API.

I would leave this open for now, maybe others can comment and we can see. Perhaps @arvidn has an idea?

For now I'll implement a derived class and forward the callback setter functions and bind underneath. Not ideal but it will do.

mfontanini:
You do realize you're wanting to change the API of a third party library just because you don't want to call `bind`? Also, are you using free functions for this? Otherwise you'll always have to bind/lambdas. What do you need to do in the callback that requires having access to the consumer but not to anything else? (I assume you're not just using global variables).

And in both asio and cppkafka it's the same, the parameters are self sufficient for whatever they do: in, say, asio's socket read callback you have access to the data read. In cppkafka's assignment callback you have access to the assignment information. Any other extra information you need has to be taken from the socket (asio) and the consumer (cppkafka). So I don't see your point.

arvidn:
perhaps I'm missing part of the context here. Is it that the underlying `librdkafka` provides some parameter that isn't forwarded to the cppkafka callback?

I would say the C++ way is to do it like it is now, which is also the way boost.asio does it. If we're talking about runtime cost of allocating space for another argument, that cost is already paid right now by using `std::function`. boost.asio takes handlers as templates and support custom allocators for them.

mfontanini:
I think the main argument is that @accelerated wants to always receive the consumer the callback is set on without having to manually bind it. I think this is fine the way it is because you can currently bind your consumer on the callback if that's what you want.

accelerated:
Ok, `bind` it is!  Next time I will run my ideas by the author before making code changes so we have a better chance of hitting the mark! Tanks everyone for the comments, good discussion overall!

PR #55, opened 2018-04-20
Print offset when dumping partition object

accelerated:
Print offset as part of the partition stream out. If offset is invalid print `#`

### DIFF
--- a/src/topic_partition.cpp
+++ b/src/topic_partition.cpp
@@ -33,6 +33,7 @@
 #include "topic_partition.h"
 
 using std::string;
+using std::to_string;
 using std::ostream;
 using std::tie;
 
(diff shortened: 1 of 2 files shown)


mfontanini:
There's a test that needs to be fixed for this (check travis). I'm all for changing this, I had actually been thinking it's annoying when you want the offset and it's not there. But I'm not sure about the formatting, I think having `topic[partition:offset]` is too compact and it may make it slightly harder to differentiate between different entries if you're printing a `TopicPartitionList` (which is at least in my experience the most common case). Maybe something like `topic[partition] = offset`? Maybe not even having the ` = offset` part when it's invalid to shorten the output? Not sure what would look best here.

src/topic_partition.cpp
```diff
@@ -92,7 +92,10 @@ bool TopicPartition::operator!=(const TopicPartition& rhs) const {
 }
 
 ostream& operator<<(ostream& output, const TopicPartition& rhs) {
-    return output << rhs.get_topic() << "[" << rhs.get_partition() << "]";
+    return output << rhs.get_topic() << "["
+                  << rhs.get_partition() << ":"
+                  << (rhs.get_offset() == RD_KAFKA_OFFSET_INVALID ?  "#" : std::to_string(rhs.get_offset()))
```
mfontanini: There's an extra space here. Also, can you add a `using std::string` clause at the top and just use `to_string` here?

accelerated:
@mfontanini done. Ready for final review.

PR #59, opened 2018-04-23
auto-gen main header file

accelerated:
Generate a single include header file for all cppkafka library:  `include/cppkafka/cppkafka.h`

File output after cmake has run:
```
/*
 * Copyright (c) 2017, Matias Fontanini
 * All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions are
 * met:
 *
 * * Redistributions of source code must retain the above copyright
 *   notice, this list of conditions and the following disclaimer.
 * * Redistributions in binary form must reproduce the above
 *   copyright notice, this list of conditions and the following disclaimer
 *   in the documentation and/or other materials provided with the
 *   distribution.
 *
 * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
 * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
 * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
 * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
 * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *
 */

#ifndef CPPKAFKA_H
#define CPPKAFKA_H

#include <cppkafka/buffer.h>
#include <cppkafka/clonable_ptr.h>
#include <cppkafka/configuration_base.h>
#include <cppkafka/configuration_option.h>
#include <cppkafka/error.h>
#include <cppkafka/exceptions.h>
#include <cppkafka/group_information.h>
#include <cppkafka/kafka_handle_base.h>
#include <cppkafka/macros.h>
#include <cppkafka/message.h>
#include <cppkafka/message_builder.h>
#include <cppkafka/metadata.h>
#include <cppkafka/producer.h>
#include <cppkafka/topic.h>
#include <cppkafka/topic_configuration.h>
#include <cppkafka/topic_partition.h>
#include <cppkafka/topic_partition_list.h>
#include <cppkafka/utils/backoff_committer.h>
#include <cppkafka/utils/backoff_performer.h>
#include <cppkafka/utils/buffered_producer.h>
#include <cppkafka/utils/compacted_topic_processor.h>
#include <cppkafka/utils/consumer_dispatcher.h>

#endif
```

NOTE: headers are sorted by default via GLOB

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ _cppkafka_'s API is simple to use. For example, this code creates a producer tha
 into some partition:
 
 ```c++
-#include <cppkafka/producer.h>
+#include <cppkafka/cppkafka.h>
 
 using namespace std;
 using namespace cppkafka;
--- a/include/cppkafka/CMakeLists.txt
+++ b/include/cppkafka/CMakeLists.txt
@@ -1,3 +1,21 @@
+# Local function to auto-generate main cppkafka.h header file
+function(make_cppkafka_header)
+    set(CPPKAFKA_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/cppkafka.h)
+    file(GLOB INCLUDE_HEADERS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "*.h" "utils/*.h")
+    foreach(header ${INCLUDE_HEADERS})
+        if (NOT ${header} MATCHES "cppkafka.h")
+            SET(CPPKAFKA_HEADERS "${CPPKAFKA_HEADERS}#include <cppkafka/${header}>\n")
+        endif()
+    endforeach()
+
+    #create file from template
+    configure_file(${PROJECT_SOURCE_DIR}/cppkafka.h.in ${CPPKAFKA_HEADER})
+endfunction()
+
+# Run file generation function
+make_cppkafka_header()
+
+# Install headers including the auto-generated cppkafka.h
 file(GLOB INCLUDE_FILES "*.h")
 file(GLOB UTILS_INCLUDE_FILES "utils/*.h")
 install(
--- a/cppkafka.h.in
+++ b/cppkafka.h.in
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2017, Matias Fontanini
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ * * Redistributions in binary form must reproduce the above
+ *   copyright notice, this list of conditions and the following disclaimer
+ *   in the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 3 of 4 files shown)


include/cppkafka/CMakeLists.txt
```diff
@@ -1,3 +1,17 @@
+# Local function to auto-generate main cppkafka.h header file
+function(make_cppkafka_header)
+    set(CPPKAFKA_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/cppkafka.h)
+    execute_process(COMMAND rm -f ${CPPKAFKA_HEADER})
```
mfontanini: I don't think this is portable. I think a better way to do this would be to have a default template file that has possibly the copyright and then a variable that you can set from here that will contain the block of includes.

We've done something similar in [another project](https://github.com/mfontanini/libtins/blob/master/include/tins/config.h.in), the file is generated [here](https://github.com/mfontanini/libtins/blob/master/CMakeLists.txt#L229).

include/cppkafka/CMakeLists.txt
```diff
@@ -1,3 +1,17 @@
...
+    file(GLOB INCLUDE_HEADERS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "*.h" "utils/*.h")
+    foreach(header ${INCLUDE_HEADERS})
+        file(APPEND ${CPPKAFKA_HEADER} "#include<cppkafka/${header}>\n")
```
mfontanini: I would add a space here, it should look like `#include <...>`.

README.md
```diff
@@ -95,7 +95,7 @@ cmake .. -DCPPKAFKA_BUILD_SHARED=0
 
 # Using
 
-If you want to use _cppkafka_, you'll need to link your application with:
+If you want to use _cppkafka_, you'll need to include `cppkafka.h` header and link your application with:
```
mfontanini: I'm not sure this clarification is needed. This is meant for users to know what they need to link to use the library. I would expect anyone who uses this C++ library knows they have to include some header to do so. Moreover, saying you "_need_ to include cppkafka.h" is not really accurate, you could include the individual headers instead.

accelerated:
Good point about the portability! I changed it to `cmake -E remove` which is portable. Fixed the header space and removed comment.

accelerated:
Done. Added the main header file and removed the superfluous line in the CMakeLists.txt.

mfontanini:
Btw, it looks like `GLOB` doesn't necessarily sort the entries. This was probably the case in whatever filesystem you are using but not on mine so I added an explicit sort call.

Also, did you manually remove an extra line before `#endif`? I was surprised I didn't see it in your PR and now that I pulled the code and ran cmake it actually generated the file with the extra new line, as you're adding `#include <....>\n` every time. Just fixed that one as well.

accelerated:
Yeah I noticed the extra line at the end after I checked-in and was going to fix it today but you beat me to it. Thanks for looking into that. I use linux, so on my system the globbing was sorted. Anyway, this was a good PR!

accelerated:
Btw, totally unrelated question, I wonder if you noticed in your librdkafka testing that when a consumer reads from multiple partitions, it does not read in round-robin order. Rather it empties the partitions one by one which is a very undesirable behavior. I ran some very simple tests, small number of messages, low throughput, even with the producer stopped and I still get this behavior. Any ideas?

mfontanini:
You should give him a chance to reply, I think he's on another timezone. I've noticed this as well. I _think_ if the backlog on the partitions is large enough you do see round robin.

PR #61, opened 2018-04-25
Log error in case consumer destructor throws

accelerated:


include/cppkafka/configuration.h
```diff
@@ -45,6 +45,8 @@
 
 namespace cppkafka {
 
+constexpr const char* library_name = "cppkafka";
```
mfontanini: Can you just move this into `Consumer::~Consumer`? This is kind of out of place up here.

src/consumer.cpp
```diff
@@ -26,7 +26,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  */
-
+#include <syslog.h>
```
mfontanini: This header doesn't exist on Windows. I'm not sure how rdkafka does this so it works on all platforms but this should do the same.

src/consumer.cpp
```diff
@@ -73,9 +73,14 @@ Consumer::~Consumer() {
         rebalance_error_callback_ = nullptr;
         close();
     }
-    catch (const Exception&) {
-        // If close throws just silently ignore until there's some
-        // logging facility (if any)
+    catch (const Exception& ex) {
+        auto& callback = get_configuration().get_log_callback();
```
mfontanini: Can you make this `const auto&`? I prefer to explicitly indicate something is const instead of expecting the user to know that's the case.

src/consumer.cpp
```diff
@@ -73,9 +73,14 @@ Consumer::~Consumer() {
...
+        if (callback) {
+            callback(*this, LOG_ERR, library_name, ex.what());
```
mfontanini: Maybe this should actually add some extra information here? I'm not sure the exception message would be enough for the user to realize what happened. Maybe something along the lines of `"Failed to close consumer: <error-message>"`?

accelerated:
Ok will address all these.

accelerated:
i'll double check for windows compatibility next time...I keep forgetting :)

include/cppkafka/exceptions.h
```diff
@@ -38,6 +38,18 @@
 
 namespace cppkafka {
 
+// Based on syslog.h levels
+enum LogLevel {
```
mfontanini: Maybe this belongs on a new `cppkafka/logging.h` file? Not sure what else would go there but I think it's slightly off in this `exceptions.h`. Also, I'd use `enum class` so this is better scoped.

src/consumer.cpp
```diff
@@ -73,9 +73,17 @@ Consumer::~Consumer() {
         rebalance_error_callback_ = nullptr;
         close();
     }
-    catch (const Exception&) {
-        // If close throws just silently ignore until there's some
-        // logging facility (if any)
+    catch (const Exception& ex) {
+        constexpr const char* library_name = "cppkafka";
+        std::ostringstream error_msg;
```
mfontanini: Minor: use `using std::ostringstream` after `include` directives and remove the `std::` specifier here.

include/cppkafka/exceptions.h
accelerated: Ok. I didn't use the class as you are not using it anywhere else, but I agree it's a better approach. Unfortunately this means that convertion enum->int has to be cast now.

mfontanini: Looks like it's being used on half on the cases. This could be cleaned up in some major version.

Issue #62, opened 2018-04-25
Feature proposal for individual partition round-robin queue polling for consumers


accelerated:
Hi @mfontanini, so in the light of what I discovered in reference to the behavior of the global queue forwarding (see comments [here](https://github.com/edenhill/librdkafka/issues/1395#issuecomment-384058260)), there is definitely a need for round-robin polling on the partitions instead of the default behavior which is to consume one batch from partition A, followed by another batch from partition B, etc. This might be OK with some applications but it's _not_ for my case and it's quite detrimental if not done properly. So the proposal would be to support both types of consume scenarios in your library. 

1) Enabling the feature could be done by:
* Provide some setter in the consumer class which would tell which polling mechanism to use (could turn on/off the feature at will). The setter would call `rd_kafka_poll_set_consumer(NULL)` if it's disabled. 
* Add a new constructor with an additional parameter which would only enable/disable the type of polling once per object. In this case we simply would not call `rd_kafka_poll_set_consumer(handle)`.
* Add a default parameter to the existing constructor which might break ABI (not desired).
* Add a cppkafka-specific configuration option for consumers which could be passed as part of the current constructor `Consumer(Configuration config)`. This could be the least invasive. 

2) Adding a `Queue` wrapper class to manage the queue handle returned by `rd_kafka_queue_get_partition`. The destructor would call the appropriate destroy function in librdkafka. This could be implemented in a similar way to your `Topic::HandlePtr` unique ptr with a destructor template argument.

3) Changing polling method by:
* keep existing `Consumer::poll()` function and call `rd_kafka_consumer_poll()` if partition polling is not enabled. 

    If enabled do:
- call `rd_kafka_poll()` to get general events.
- for each topic partition in the topic partition list (a copy of the topic partition list will be kept during the call to `assign` - see 4) call `rd_kafka_consume_queue()`. Note that only one partition will be consumed on each call to `Consumer::poll()` as a static round-robin counter will be kept internally to the function. 
- return the message if any (_or even better, poll until one valid message is returned or all partitions have been polled_)

  3.a) Alternatively keep `poll()` unchanged and add `poll_partitions()` function however in this case some error must be thrown if the user enables partition polling and then calls `poll()`.

[...]

PR #63, opened 2018-04-27
round robin polling for assigned partitions

accelerated:
Created two polling strategies to the consumer class - the default one represents the current behavior. The other polling strategy (aka partition round-robin) allows for individual consumption of assigned partitions instead of the batch method (current implementation with common forward queue). 
Also created a Queue helper class to better separate polling logic and to generalize the polling mechanism at the Consumer layer i.e. the Consumer now consumes from a list of queues (may contain a single queue in the common forwarding case) irrespective of the polling strategy and is as such decoupled from underlying queue logic.


accelerated:
Initially in my Consumer::poll() method I was thinking to poll all queues until I get a valid message (see below) but the issue with this is the blocking call on `rd_kafka_poll`. Since I have to respect the `timeout` argument, I either have the choice to wait in the event poll or in the message consume functions. The problem with spending all the timeout in the `rd_kafka_poll` call is that events represent a small percentage of total events, the majority being messages, so it's much better to block on `consume`. So I could have moved the `rd_kafka_poll` call at the end with a timeout of 0, but then the problem is how to split the timout into N queues? So finally I opted to poll a single partition per `Consumer::poll()` call. 

```
Message Consumer::poll(milliseconds timeout) {
    // Poll the general event queue
    rd_kafka_poll(get_handle(), static_cast<int>(timeout.count()));

    Message message;
    size_t polls = partition_queues_.size();
    while (polls--) {
        message = partition_queues_[queue_index_++].consume(milliseconds(0));
        if (message) {
            break; //got a valid message
        }
        // adjust the queue index
        if (queue_index_ >= partition_queues_.size()) {
            queue_index_ = 0;
        }
    }
    return message;
}
```

A similar problem happens in the batch poll case. How do I split batches between N partitions? Do I get `batch_size/N` messages from each queue? what if some queues are empty? Then I would potentially return an incomplete batch set, while some other queues still contain messages. Also the poll timeout issue applies here as well. So once again I decided to only batch each partition on each call..it makes the logic much simpler and I also leave the `rd_kafka_poll` at the end, so that most of the blocking is done waiting for messages. 

Thoughts on this?

accelerated:
After testing, in the default case i.e. POLL_COMMON_QUEUE, everything works as before and all messages and events trigger. However in the round-robin case, it seems that `rd_kafka_poll` is not working and since i'm relying on partition assignment event to create my queue list, nothing works. I can't really figure out what I'm missing...perhaps call `rd_kafka_consumer_poll_start`?

accelerated:
This final version works well. I fixed all the issues related to rd_kafka_poll not working.

mfontanini:
As I mentioned in #62, other users may want to consume using other strategies. Instead of tying this to the `Consumer` class, I think this should be an external utility (e.g. `RoundRobinConsumerHelper` or something like that).

The reason why I created different utility classes (e.g. `ConsumerHelper`, `BackoffCommitter`, `BufferedProducer`) is because I think the `Consumer`/`Producer` classes should be kept simple: they're basically just a wrapper over the kafka handle. I think behavior like this one belongs on a separate class, maybe even a hierarchy which allows plugin in a poll strategy. If tomorrow I want to implement some sort of consumption (e.g. sorted by time), it would be nice if I could do it by simply inheriting a helper class and define some function that chooses which queue to consume from next and the rest is done automatically. Tying a "round robin strategy" on `Consumer` could be chaotic as in the future if 10 different strategies are implemented, the `Consumer` class will become really bloated and complex.

accelerated:
Would you lean more towards inheritance, i.e. some modified consumer derived class which hides the poll methods of the parent class or a helper class where you pass-in a consumer object and then it queries queues on it and polls them? I would lean more towards the latter

accelerated:
Or thirdly, you could inject a polling strategy  object into the consumer which would alter its runtime poll functions

mfontanini:
I meant just like the `ConsumerHelper` class works: get a `Consumer` and work with it. The `Consumer` class shouldn't be changed unless absolutely required. The `CompactedTopicProcessor` class currently wraps whatever assignment/revocation callbacks the consumer has and injects a proxy that adds some extra logic.

accelerated:
Ok thanks, will submit final PR soon.

accelerated:
@mfontanini Code complete, please review. Will fix the failing assertion...

mfontanini:
Please don't use `git reset` and `git push -f`. It's really hard to see the deltas of your changes if there's always just a single commit that contains all the changes, especially if you push while I'm reviewing the code (but also for any changes you have to apply after the review).

accelerated:
Ok done. I missed the scenario where the topic partitions list could contain multiple (different) topics during assignment/revocation.

include/cppkafka/utils/roundrobin_poll_adapter.h
```diff
@@ -173,25 +174,39 @@ class RoundRobinPollAdapter
      */
     MessageList poll_batch(size_t max_batch_size, std::chrono::milliseconds timeout);
     
+    /**
+     * \brief Gets the number of assigned partitions that can be polled across all topics
+     *
+     * \return The number of partitions
+     */
+    size_t get_num_partitions();
```
mfontanini: Wouldn't it be better to allow getting the current assignment? That is, the list of all topic+partitions.

include/cppkafka/utils/roundrobin_poll_adapter.h
```diff
@@ -173,25 +174,39 @@ class RoundRobinPollAdapter
...
 private:
     class CircularBuffer {
-        using qlist = std::list<Queue>;
-        using qiter = qlist::iterator;
     public:
-        qlist& ref() { return queues_; }
+        // typedefs
+        using toppar_t = std::pair<std::string, int>; //<topic, partition>
+        using qmap_t = std::map<toppar_t, Queue>;
+        using qiter_t = qmap_t::iterator;
```
mfontanini: Is this type alias really necessary? The type is only used once and only internally.

include/cppkafka/utils/roundrobin_poll_adapter.h
```diff
@@ -173,25 +174,39 @@ class RoundRobinPollAdapter
...
+    
 private:
     class CircularBuffer {
-        using qlist = std::list<Queue>;
-        using qiter = qlist::iterator;
     public:
-        qlist& ref() { return queues_; }
+        // typedefs
+        using toppar_t = std::pair<std::string, int>; //<topic, partition>
+        using qmap_t = std::map<toppar_t, Queue>;
```
mfontanini: I would call this `QueueMap`, the name is short enough that it's not worth abbreviating.

include/cppkafka/utils/roundrobin_poll_adapter.h
```diff
@@ -173,25 +174,39 @@ class RoundRobinPollAdapter
...
+        using qiter_t = qmap_t::iterator;
+        
+        qmap_t& ref() {
```
mfontanini: I would call this `get_queues` or something like that. Again, no need for extreme abbreviations for names this short.

include/cppkafka/utils/roundrobin_poll_adapter.h
```diff
@@ -173,25 +174,39 @@ class RoundRobinPollAdapter
...
+            return queues_;
+        }
+        
         Queue& next() {
```
mfontanini: `get_next_queue` here maybe?

src/utils/roundrobin_poll_adapter.cpp
```diff
@@ -0,0 +1,182 @@
...
+    // take over the revocation callback
+    consumer_.set_revocation_callback([this](const TopicPartitionList& partitions) {
+        on_revocation(partitions);
+    });
+    // take over the rebalance error callback
+    consumer_.set_rebalance_error_callback([this](Error error) {
+        on_rebalance_error(error);
+    });
+    // make sure we don't have any active subscriptions
+    if (!consumer_.get_subscription().empty()) {
```
mfontanini: This should be checked before setting all the callback, otherwise you'll leave the `Consumer` dirty.

src/utils/roundrobin_poll_adapter.cpp
```diff
@@ -0,0 +1,182 @@
...
+    if ((remaining_count <= 0) || empty_list) {
+        // the entire batch was filled
+        return messages;
+    }
+    // batch from the next partition
+    MessageList partition_messages = partition_queues_.next().consume_batch(remaining_count, timeout);
+    if (messages.empty()) {
+        return partition_messages;
+    }
+    if (partition_messages.empty()) {
```
mfontanini: This can be removed. If `partition_messages` is empty, the lines below won't do anything anyway.

include/cppkafka/utils/roundrobin_poll_adapter.h
accelerated: My initial intention is to eventually allow polling for all (default) or certain topics only by adding a setter function. In the case of certain topics, then this function will not be identical to get current assignment.

src/CMakeLists.txt
```diff
@@ -1,24 +1,4 @@
...
-    error.cpp
-
-    kafka_handle_base.cpp
-    producer.cpp
-    consumer.cpp
-
-    utils/backoff_performer.cpp
-    utils/backoff_committer.cpp
-)
+file(GLOB SOURCES *.cpp utils/*.cpp)
```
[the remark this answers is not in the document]
accelerated: Could you be more specific as to the _bad side effects_? I've been using this as long as I can remember and never had problems. Also we're using it for generating the global header file which is why I wanted to add it here. I also don't like having to edit the `CMakeLists.txt` file every time a cpp file gets added.

src/consumer.cpp
```diff
@@ -248,6 +246,24 @@ vector<Message> Consumer::poll_batch(size_t max_batch_size, milliseconds timeout
...
+    queue.disable_queue_forwarding();
+    return queue;
+}
+
+Queue Consumer::get_consumer_queue() const {
+    return Queue::make_non_owning(rd_kafka_queue_get_consumer(get_handle()));
+}
+
+Queue Consumer::get_partition_queue(const TopicPartition& partition) const {
+    Queue queue = Queue::make_non_owning(rd_kafka_queue_get_partition(get_handle(),
```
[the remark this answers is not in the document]
accelerated: I thought so too based on the librdkafka documentation (misleading _remove reference_ comment). However based on my initial testing, once you destroy a queue, it's gone forever for the lifetime of the consumer. And since I need to reuse them, I can't destroy them. I will redo this test once again to confirm but unfortunately it's how the library works.

src/utils/roundrobin_poll_adapter.cpp
accelerated: I changed the logic...instead of rejecting, I get current assignments which is the desired behavior.

src/utils/roundrobin_poll_adapter.cpp
```diff
@@ -0,0 +1,161 @@
...
+        return messages;
+    }
+    // concatenate both lists
+    messages.reserve(messages.size() + partition_messages.size());
```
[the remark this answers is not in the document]
accelerated: I use reserve for performance reasons and I think we should keep it as the batch size could be quite large. That's the reason I added the return line above the reserve, in case it's empty there's not need to go forward.

src/utils/roundrobin_poll_adapter.cpp
accelerated: This was added to prevent the `reserve()`

src/utils/roundrobin_poll_adapter.cpp
```diff
@@ -0,0 +1,161 @@
...
+    // batch from the group event queue first
+    MessageList messages = consumer_queue_.consume_batch(remaining_count,
+                                                         empty_list ? timeout : milliseconds(0));
+    remaining_count -= messages.size();
+    if ((remaining_count <= 0) || empty_list) {
+        // the entire batch was filled
+        return messages;
+    }
+    // batch from the next partition
+    MessageList partition_messages = partition_queues_.next().consume_batch(remaining_count, timeout);
```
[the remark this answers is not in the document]
accelerated: I don't timeout twice, the first time checks if the partition number is 0 and in that case only applies the timeout for consumer_queue, otherwise it's 0. 
```
MessageList messages = consumer_queue_.consume_batch(remaining_count,
                                                         empty_list ? timeout : milliseconds(0)); ------>> here
```

Regarding the batching across multiple partitions, I've though about that...not really sure what's best. I felt I should have the same behavior as the single `poll` which only polls one partition per call. It also becomes complicated with the timeout paramter...when you poll multiple partitions under the hood, where do you apply the timeout? On which queue? 
The reason why I added the `get_num_pollable_partitions()` is so that users can poll all partitions until a valid message is found, instead of polling N times with a timeout. In this case you timeout once. 

Example:
```
//current code 
while (1) {
    Message msg = adapter->poll(timeout); //if you have N partitions you get N*timeout potentially
    if (msg) ...                  }

// better alternative (polls all partitions until a msg is found)
while (1) {
    int num = adapter->get_num_pollable_partitions();
    while (num--) {
           Message msg = adapter->poll(); //no timeout
           if (msg) break; 
    }
    std::this_thread::sleep_for(timeout);
}
```

Of course, I could implement this inside `poll` itself, which is fine, but then `poll_batch()` will not work in the same way. Ideally `poll_batch` would poll `max_batch/num_partitions` from each partition so that it's equally distributed. Again, not sure that's best...if you have a preference, let me know.

src/utils/roundrobin_poll_adapter.cpp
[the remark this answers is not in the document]
mfontanini: Insert already expands the vector (once) so this call is useless. `insert` knows how many elements there are between the two you provide so it knows how much to enlarge the vector.

src/utils/roundrobin_poll_adapter.cpp
mfontanini: See my comment below.

src/utils/roundrobin_poll_adapter.cpp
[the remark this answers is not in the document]
mfontanini: > I don't timeout twice

Fair enough. Does calling poll with timeout 0 work as expected? I imagine so but just checking.

> where do you apply the timeout

You could keep track of how much time it took you to process each queue and decrement that from the timeout. This could actually be applied to the current code, but if it works as it is, that works. Let's keep this simple for now.

src/consumer.cpp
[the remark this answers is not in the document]
mfontanini: Looks like this is worth raising as a bug on the rdkafka repo. I don't think that should be the expected behavior. I'm wondering if this leaks memory as it is though, it seems kind of eerie.

include/cppkafka/utils/roundrobin_poll_adapter.h
mfontanini: I mean, you always have a list of queues you're pulling from. Whatever your list of queues is should be accessible from the outside, regardless of whether they came through a partition assignment or it was the user who set them.

src/utils/roundrobin_poll_adapter.cpp
mfontanini: I don't understand. Are you referring to the current code or some uncommitted fix?

src/utils/roundrobin_poll_adapter.cpp
accelerated: Ok fixed.

mfontanini:
There should have been a test for `commit` with no topic/partitions as the parameter (my bad, should have asked for it). What happens when you call commit with null?

Also, there should be some tests for this so we can guarantee it works. You can have a look at how the [travis build](https://github.com/mfontanini/cppkafka/blob/master/.travis.yml) sets up kafka to make tests work. There basically need to be 2 topics (cppkafka_test1 and cppkafka_test2) with 3 partitions each, and when you invoke cmake you should pass `-DKAFKA_TEST_INSTANCE=<endpoint where kafka is>`.

src/utils/roundrobin_poll_adapter.cpp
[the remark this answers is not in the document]
mfontanini: If no messages ever arrive when you're blocked, won't you stop sending heartbeats (because you're not polling the consumer queue), hence a rebalance will occur?

src/consumer.cpp
```diff
@@ -223,27 +223,21 @@ Message Consumer::poll(milliseconds timeout) {
...
+        return MessageList();
     }
-    vector<Message> output;
-    raw_messages.resize(result);
-    output.reserve(result);
-    for (const auto ptr : raw_messages) {
-        output.emplace_back(ptr);
-    }
-    return output;
+    return MessageList(raw_messages.begin(), raw_messages.end());
```
mfontanini: Nice, looks much better.

accelerated:
Will add the travis tests...stay tuned.

mfontanini:
Keep in mind you can run this locally by running `./tests/cppkafka_tests`. 

Regarding the commit part, it could at least check if something is actually committed (see the `consumer offset commit` test in `consumer_test.cpp`).

src/utils/roundrobin_poll_adapter.cpp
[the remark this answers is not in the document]
accelerated: I guess instead of blocking on one of the queues, I could block on the consumer queue, which will ensure that at least heartbeats are sent ?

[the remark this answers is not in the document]
mfontanini: I would just not block. As soon as you hit an empty queue you'll hit the timeout anyway so I don't think there's a need to do that.

accelerated:
My suspicions with destroying the queues after use are confirmed. librdkafka will flush and destroy all messages and the queue instance, however if you query a new reference to this queue at a later point in time, a new queue is created yet it's not functional - i.e. no messages are enqueued and no messages are fetched from the broker - same happens with the global consumer queue which stops working completely. 
I believe that one must call `rd_kafka_consume_start_queue()` with a specific offset but even that fails for me and if I then poll the main event queue, I get a crash. I've tried a bunch of different combinations of `rd_kafka_consume_stop_queue` then destroy and start...and nothing works. 
The current implementation with non-owning works very well, I even tested polling via one adapter, deleting it, polling directly via consumer and then polling again via a new adapter and all works fine and the queues transition well even during a rebalance. Haven't seen duplicate messages being consumed either which means offsets are properly managed. I will add Travis test cases asap.

mfontanini:
Could you create a ticket on rdkafka regarding this with your findings? I'm not doubting what you did but this looks very broken on the rdkafka side and it shouldn't be happening. At some point the queues should be destroyed and I don't know what happens if you don't do so.

accelerated:
See opened [ticket](https://github.com/edenhill/librdkafka/issues/1792) with librdkafka.

accelerated:
Hi @mfontanini, so I'm looking into building some test cases...would you be open to templetizing the `ConsumerRunner` class? I've created a derived class from `Consumer` called `RoundRobinConsumer`, which contains a `RoundRobinPollAdapter` and all the poll functions are delegated via the adapter, so now I can use `BasicConsumerDispatcher<RoundRobinConsumer>`. However the `ConsumerRunner` is directly using `ConsumerDispatcher` class and it would be nice if we could templetize it. You also have a hard-coded consumer_ member so I can't really override that...unless, instead of templetizing the class I can add another constructor where I can inject the dispatcher, and in the default constructor case, I would inject the ConsumerDispatcher...so as not to change the ABI of your test class.

mfontanini:
`Consumer` is not really meant to be inherited, why do you need that class? The point of the adapter you created is it runs on top of a consumer and you don't need to touch the consumer class at all.

Can't you test it like the "consume batch"  or "consumer throttle"? The `ConsumerRunner` is mostly used to start consuming from the topics in a clean state, as you don't know if any previous tests left any uncommitted messages in them (see [this test](https://github.com/mfontanini/cppkafka/blob/master/tests/consumer_test.cpp#L219)). The class also allow you to fetch messages but that's not necessarily the only use use case for it.

accelerated:
Well although Consumer is not meant to be inherited, i actually don't need any polymorphic behavior so that's fine with me.   The thing is, I do need the ConsumerDispathcher for other applications as well outside of testing, so I would like to use it jointly with the round robin poll adapter as it's a nice/useful class to have. There's two ways that i can think of:
1) derive from it which is ok since it's always accessed from the derived-most class in the ConsumerDispatcher. The poll functions will hide the base implementations, and the other methods are inherited. Works perfectly.
2) convert the round robin adapter into a real PIMPL idiom class, and forward all calls to the internal consumer reference. It's more coding but it avoids inheritance and works just as well with the ConsumerDispatcher. On the down side, any new method added to Consumer will have to be mirrored in the adapter, which you get for free if you implement 1)

Let me know which one you prefer.

The ConsumerRunner is a different story. I want to make use of it because it's easy to test with. Also you stop the runner when the number of eof's is >= to your partitions, which means that if i run my tests AFTER this one, and they also use the runner class, each time the kafka queues will have the latest offsets. I don't really need more or a clean state than this. 

Anyway, that's my justification. I can implement 2) since you seem to dislike the inheritance model (although simpler) and I can add the RoundRobinConsumer to the adapter header file. As for the runner...i dunno, if you dont want to templetize it i'll think of something else, but why reinvent the wheel? ;)

mfontanini:
I see what your issue is now. So going back to what I mentioned a few messages back, it would be really nice if the algorithm to choose which queue to consume from could be user provided so you don't have to basically reimplement the entire `RoundRobinPollAdapter` from scratch except instead of doing `queue_iterator++` every time, you do something else. The round robin strategy would just be one cppkafka provides but users could define their own.

Then going back to the round robin consumer, if that's a generic class by using the above, it could work. I imagine the round robin strategy would have to be a pimpl, so it's simple to plug in the strategy. As for inheritance, I'm not against it at all. I just don't think it's great to use inheritance and override methods which aren't virtual. I know it works fine but it just feels like you're hacking your way around the class.

I'm not sure how to allow this to work with the `ConsumerDispatcher` without introducing a new class. One way I can think is to have the dispatcher accept another callback in `run` that is basically the "polling" functor. By default it would do `consumer.poll()` but you could set one that's a lambda which calls `poll` on the adapter/helper that actually implements the consumption.

accelerated:
Hi @mfontanini, let's take a step back and for a second pretend this is an ideal world where we _don't_ care about ABI compatibility (but we _care_ about preserving compile-time compatibility). 

1) In this ideal world, consumers and producers would be fully polymorphic implementing pure abstract interfaces, allowing people to build their own versions and at the same time still use a plethora of util classes which would interact through these pure abstract interfaces. It would be perfect decoupling and it would be easy to create lots of new behaviors for these classes. C++ at its best. But we can't do that because of ABI.

2) In an _almost_ ideal world, the consumers would have virtual poll functions and virtual produce functions and at least people can still derive from these classes and implement their own poll/produce behaviors. These new implementations would still work well with util classes who would interact through base Consumer and Producer classes. Not as clean as the pure abstract interfaces from the _ideal world_ but not terrible either. Also both 1) and 2) don't require the exposure of the Queue getter APIs.

Now let's look at the present _not-so-ideal_ world that we have. 

3) One way to allow run-time behavioral changes to the Consumer (and perhaps Producer) class is to implement the _behavior pattern_ which is well known. In this pattern, we inject a `PollStrategy` (and if you like a `ProduceStrategy` for the producer) into the consumer. If nothing gets injected, the consumer will have a `DefaultPollStrategy` which contains the current logic. The PollStrategy would be a pure abstract interface with various `poll` functions (you can add more later such as poll by timestamp, etc) and users can create their own PollStrategy implementations: round-robin, poll-till-eof, poll-in-special-batches, poll-by-timestamp, etc. All they need to do is inject this into the Consumer class via `Consumer::set_poll_strategy()` and things would work as before. The good part about this is that you don't have to expose any Queue getters and people can still make use of all the util classes as before. Also the util classes don't even have to be changed as you are always working with a Consumer instance. No polymorphism or virtualization required and it conserves ABI perfectly. Internally, all the poll functions will just forward the call to the `PollStrategy` object in place. 
This was what I proposed way back but you turned it down because you said the Consumer class was meant to be a very simple wrapper and you didn't want to complicate its behavior. Fine. I think it's an unfortunate decision but it's your code and so be it. 

4) Comes the PIMPL variant (or the _adapter pattern_ if you like) which is a second-class citizen IMHO, where `poll` methods are overriden, and other methods are forwarded directly to the internal Consumer object held. Another way of doing this is the hybrid inheritance which I proposed (for testing), not that it's any better, but requires less typing. This PIMPL/adapter implementation involves no ABI changes and no changes to the consumer class, however people can't benefit from using util classes, doesn't scale too well and it's hard to couple with other adapters. In 1), 2) and 3) you can create various strategies or various derived classes that can derive from others, etc...and thus chain functionality. With 4) overloads cannot be intermixed. 
Luckily the `BasicConsumerDispatcher` is a template class (wonder how that happened but it's very fortunate), so at least I can use the PIMPL adapter there. 

As the owner for this library, you can do:

* Make one big ABI breaking change (won't require compile changes) and we implement 1) and never look back. Sometimes it's good to take a bold stride forward.
* Keep the old Consumer and Producer but create new classes called KafkaConsumer/KafkaProducer which implement 1) and maybe add additional APIs to the util classes to use these interfaces instead. This won't break ABI and it opens up the library. The current Consumer and Producer classes will slowly wither away and die as people stop using them (small tear).
* Stick to 3) or worse 4) and we see. I can convert the RoundRobinPollAdapter to be full PIMPL compliant (not just the poll functions), use the `BasicConsumerDispatcher<RoundRobinPollAdapter>` for testing and in my daily code. And I can write code similar to ConsumerRunner for testing purposes only (if you don't want to templetize it). I belive that ConsumerRunner can easily use the RoundRobinPollAdapter because it always consumes till EOF on all partitions. 

Now to answer your questions:
> I see what your issue is now. So going back to what I mentioned a few messages back, it would be really nice if the algorithm to choose which queue to consume from could be user provided so you don't have to basically reimplement the entire RoundRobinPollAdapter from scratch except instead of doing queue_iterator++ every time, you do something else. The round robin strategy would just be one cppkafka provides but users could define their own.

This is exactly what 3) aka the _behavior pattern_ does but in a clean way.

> Then going back to the round robin consumer, if that's a generic class by using the above, it could work. I imagine the round robin strategy would have to be a pimpl, so it's simple to plug in the strategy. As for inheritance, I'm not against it at all. I just don't think it's great to use inheritance and override methods which aren't virtual. I know it works fine but it just feels like you're hacking your way around the class.

I think what you're proposing is a strange mix of adapter and behavior patterns? Do I understand correctly?

```
class GenericAdapter {
public: 
     void set_poll_strategy(IPollStrategy* strategy) {
          poll_strategy_ = strategy;
          poll_strategy_->set_instance(consumer_);
     }

     Message poll() {
          if (poll_strategy_) poll_strategy_->poll();
          else consumer_.poll();
     }   

     //all non-poll methods are forwarded directly to consumer_.
private:
    Consumer& consumer_; //added in the constructor
    IPollStrategy*  poll_strategy_; //optional
}
```

mfontanini:
The purpose of the `Consumer` and `Producer` classes are to be pretty much zero cost abstractions on top of the rdkafka consumer/producer handles. For any more complex logic, I would move that somewhere else. Otherwise, who implements the rd_kafka_t wrapping logic? Does the consumer do that? What if you override e.g. the `poll` method? It just feels like you're taking over part of the responsibilities of that class. I think of `Consumer` as a building block with which you can create more complex logic. Again, see the `BackoffCommitter` or `BufferedProducer` templates. Those ones just use a consumer under the hood to achieve some logic that's really useful but you probably don't want to embed into the `Consumer` / `Producer` because that's out of their scope. Those are really only meant to be dumb, simple, wrappers.

In your `GenericAdapter` example above, that would work except `poll` wouldn't have an optional strategy, it would always be there (I think you mixed up my comment regarding this and the `ConsumerDispatcher` here). Then the rest of the methods would be forwarded to the underlying consumer.

accelerated:
Ok, take 3. I will rename the `RoundRobinPollAdapter` to `RoundRobinPollStrategy` as it should be a simple class doing just that and then create a `ConsumerPollAdapter` class which will implement the PIMPL idiom in a generic way and accept different strategies. Example:
```
class IPollStrategy
{
public:
    IPollStrategy(Consumer& consumer);
    virtual ~IPollStrategy() = default;
    Consumer& get_consumer() { return consumer_; } //non virtual 

    // interface methods
    virtual void set_timeout() = 0;
    virtual timeout get_timeout() = 0;
    virtual Message poll() = 0;
    virtual Message poll_batch() = 0;
    // ...

private:
    Consumer& consumer_;    
};

class RoundRobinPollStrategy : public IPollStrategy
{
public:
    using IPollStrategy::IPollStrategy;

    //implement virtual methods 
};

class ConsumerPollAdapter 
{
public: 
    PollAdapter(std::unique_ptr<IPollStrategy> strategy); //any strategy can be added here...

    // Consumer poll methods...
    Message poll() {
        strategy->poll();
    }

    // Consumer non poll methods...
    void non_poll_method() {
        strategy->get_consumer()->non_poll_method();
    }

    //Should Queue getter methods be present ??? or throw if used ???

private:
     std::unique_ptr<IPollStrategy> strategy_;
};

// ok to use this now as the adapter implements all Consumer methods
using ConsumerPollDispatcher = BasicConsumerDispatcher<ConsumerPollAdapter>;
```

Still not sure how to use the `ConsumerRunner`...

accelerated:
BTW have you given any though on how to enhance the `Consumer` and the `IPollStrategy` to accept timestamps as arguments i.e. say i need to start consuming a queue based on a messages newer than time T?? I think I may have this requirement soon...

accelerated:
> Are you planning on having the logic of getting the queues on assignment/clearing them on revocation on some place outside of the RoundRobinPollStrategy? I really think that should be put in some common place so the poll strategy doesn't need to reimplement this logic for every new poll strategy you want to implement as that is probably going to look the same.

Hmm not a bad idea, i could put that logic inside a `PollStrategyBase` class, however i wonder if others won't need different structures for storing and managing their queues? They may use tuples for storing queue + other info they might need when using the queue. Maybe i could store a `pair <queue,boost::any>` in the map so that any other metadata can fit.

accelerated:
Looking at the code, I'm not sure the generic `ConsumerPollAdapter` using PIMPL would work...there's just too many Consumer methods to forward, I totally forgot the ones present in the `KafkaHandleBase` and it seems very overkill. Also if new methods are added to the Consumer class, then the adapter will fall out of sync. Thoughts? But then w/o the full PIMPL stuff, I can't properly test...catch22 situation.

accelerated:
I added two tests but not sure why there's no message consumption happening. I have to try and recreate the Travis environment and run the tests directly. I even ran a test with the original consumer class and I still get the same result. If you can take a look quickly and see...maybe I'm not doing something right. The original test will go away...it was just to try and see if there's something wrong with the adapter.

accelerated:
Tests passed. This looks like the final version unless you have more comments.

mfontanini:
I've been quite busy recently and haven't had a chance to look at this. I'll review it in the next couple of days.

tests/roundrobin_poll_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    runner.try_join();
+    
+    // Check that we have all messages
+    REQUIRE(runner.get_messages().size() == total_messages);
+    
+    // Check that we have one message from each partition in desired order
+    vector<int> partition_order = make_roundrobin_partition_vector(total_messages);
+    
+    for (int i = 0; i < total_messages; ++i) {
+        REQUIRE(runner.get_messages()[i].get_partition() == partition_order[i+1]);
```
mfontanini: I'm probably missing something but how does this guarantee the starting topic/partition will be what you expect? If say, you somehow took slightly longer to produce messages than you expect, you could end up having the consumer poll through the first partition, jump to the second and then get a message on that one. The tests seem to have succeeded but I'm curious about that potential race condition.

tests/roundrobin_poll_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+TEST_CASE("roundrobin consumer test", "[roundrobin consumer]") {
+    TopicPartitionList assignment;
+    int messages_per_partition = 3;
+    int total_messages = KAFKA_NUM_PARTITIONS * messages_per_partition;
+
+    // Create a consumer and subscribe to the topic
+    PollStrategyAdapter consumer(make_consumer_config());
+    TopicPartitionList partitions;
+    for (int i = 0; i < KAFKA_NUM_PARTITIONS; partitions.emplace_back(KAFKA_TOPICS[0], i++));
+    consumer.assign(partitions);
```
mfontanini: Any reason why you're not using `subscribe` here? That involves a slightly trickier code flow so it may be worth using it instead.

accelerated: Ok will add subscribe.

tests/roundrobin_poll_test.cpp
accelerated: Well most likely the partitions are emtpy, which means all the EOFs arrive in sequence, and then the first poll timeout is 1s which is more then enough to publish 3 messages, especially since kafka broker is on localhost. In my testing, the broker is in a remote datacenter cluster and it still works. I have slightly changed the test now, allowing for a different start partition in case the first poll (or other ones) times out.

accelerated:
@mfontanini any chance you can merge this?

mfontanini:
Seems like this needs a rebase.

accelerated:
yeah because of other PRs which came after and were accepted before :). I'll fix it

mfontanini:
Thanks!

PR #64, opened 2018-05-13
Add yield

amirshavit:
Simply calls rd_kafka_yield; allows consumers and producers to abort the
current callback dispatcher.


mfontanini:
Thanks!

PR #66, opened 2018-05-15
Allow access to the user-supplied delivery callback.

accelerated:
**Description**:
- Currently the `buffered_producer` overrides the user-supplied delivery callback in the producer config. This PR allows that callback to be called as well. This is useful for any kind of other task the client application might be doing such as logging/reporting/permanent storage, etc.
- Added a method for querying the length of the buffer. This might be useful if application wants to flush messages after a certain buffer size without having to keep count of added messages. 

**Additional suggestion**: perhaps adding a watermark setting for the buffer size and when that watermark is reached, `flush` is called automatically.

### DIFF
--- a/include/cppkafka/producer.h
+++ b/include/cppkafka/producer.h
@@ -43,6 +43,7 @@ namespace cppkafka {
 class Topic;
 class Buffer;
 class TopicConfiguration;
+class Message;
 
 /**
  * \brief Producer class
@@ -86,39 +87,44 @@ public:
     };
 
     /**
-     * Constructs a producer using the given configuration
+     * \brief Constructs a producer using the given configuration
      *
      * \param config The configuration to use
      */
     Producer(Configuration config);
 
     /**
-     * Sets the payload policy
+     * \brief Sets the payload policy
      *
      * \param policy The payload policy to be used
      */
     void set_payload_policy(PayloadPolicy policy);
 
     /**
-     * Returns the current payload policy
+     * \brief Returns the current payload policy
      */
     PayloadPolicy get_payload_policy() const;
 
     /**
-     * Produces a message
+     * \brief Produces a message
      *
-     * \param topic The topic to write the message to
-     * \param partition The partition to write the message to
-     * \param payload The message payload
+     * \param builder The builder class used to compose a message
      */
     void produce(const MessageBuilder& builder);
+    
+    /**
+     * \brief Produces a message
+     *
+     * \param message The message to be produced
+     */
+    void produce(const Message& message);
 
     /**
      * \brief Polls on this handle
      *
      * This translates into a call to rd_kafka_poll.
      *
-     * The timeout used on this call is the one configured via Producer::set_timeout.
+     * \remark The timeout used on this call is the one configured via Producer::set_timeout.
      */
     int poll();
 
@@ -136,7 +142,7 @@ public:
      *
      * This translates into a call to rd_kafka_flush.
      *
-     * The timeout used on this call is the one configured via Producer::set_timeout.
+     * \remark The timeout used on this call is the one configured via Producer::set_timeout.
      */
     void flush();
 
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -128,6 +128,27 @@ public:
     void clear();
     
     /**
+     * \brief Sets the maximum amount of messages to be enqueued in the buffer.
+     *
+     * After 'max_buffer_size' is reached, flush() will be called automatically.
+     *
+     * \param size The max size of the internal buffer. Allowed values are:
+     *             -1 : Unlimited buffer size. Must be flushed manually (default value)
+     *              0 : Don't buffer anything. add_message() behaves like produce()
+     *            > 0 : Max number of messages before flush() is called.
+     *
+     * \remark add_message() will block when 'max_buffer_size' is reached due to flush()
+     */
+    void set_max_buffer_size(ssize_t max_buffer_size);
+    
+    /**
+     * \brief Return the maximum allowed buffer size.
+     *
+     * \return The max buffer size. A value of -1 indicates an unbounded buffer.
+     */
+    ssize_t get_max_buffer_size() const;
+    
+    /**
      * \brief Get the number of messages in the buffer
      *
      * \return The number of messages
@@ -175,6 +196,7 @@ private:
     ProduceFailureCallback produce_failure_callback_;
     size_t expected_acks_{0};
     size_t messages_acked_{0};
+    ssize_t max_buffer_size_{-1};
 };
 
 template <typename BufferType>
@@ -206,7 +228,6 @@ void BufferedProducer<BufferType>::flush() {
         produce_message(messages_.front());
         messages_.pop();
     }
-
     wait_for_acks();
 }
 
@@ -239,6 +260,19 @@ void BufferedProducer<BufferType>::clear() {
 }
 
 template <typename BufferType>
+void BufferedProducer<BufferType>::set_max_buffer_size(ssize_t max_buffer_size) {
+    if (max_buffer_size < -1) {
+        throw Exception("Invalid buffer size.");
+    }
+    max_buffer_size_ = max_buffer_size;
+}
+
+template <typename BufferType>
+ssize_t BufferedProducer<BufferType>::get_max_buffer_size() const {
+    return max_buffer_size_;
+}
+
+template <typename BufferType>
 size_t BufferedProducer<BufferType>::get_buffer_size() const {
     return messages_.size();
 }
@@ -247,7 +281,10 @@ template <typename BufferType>
 template <typename BuilderType>
 void BufferedProducer<BufferType>::do_add_message(BuilderType&& builder) {
     expected_acks_++;
-    messages_.push(std::move(builder));
+    messages_.push(std::forward<BuilderType>(builder));
+    if ((max_buffer_size_ >= 0) && (max_buffer_size_ <= (ssize_t)messages_.size())) {
+        flush();
+    }
 }
 
 template <typename BufferType>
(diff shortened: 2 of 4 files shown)


accelerated:
On a side note, is there a reason you deleted the copy and assignment Buffer operators? In both cases the passed references need to be valid for the lifetime of the builder object. 
```
        const auto& key = message.get_key();
        const auto& payload = message.get_payload();
        builder.partition(message.get_partition())
               .key(Buffer(key.get_data(), key.get_size()))
               .payload(Buffer(payload.get_data(), payload.get_size()))
               .user_data(message.get_user_data());
```
could be more nicely written as:
```
        builder.partition(message.get_partition())
               .key(message.get_key())
               .payload(message.get_payload())
               .user_data(message.get_user_data());
```

mfontanini:
`Buffer` attempts to be as safe as it can. It's a dangerous class because it points to a buffer it doesn't own so if you allow copying, you increase the odds of someone misusing it and e.g. storing the message's payload into a vector, and then it will blow up. You can still make copies (as your example shows) but you have to make them manually, which is still short but makes sure you know what you're doing.

The PR looks good except... something blew up? That's odd.

accelerated:
/home/adamian/mbig/git/cppkafka/tests/cppkafka_tests [producer]
Running test "simple production" @ /home/adamian/mbig/git/cppkafka/tests/producer_test.cpp:51
Done. 32/32 assertions succeeded in 5795ms
Running test "multiple messages" @ /home/adamian/mbig/git/cppkafka/tests/producer_test.cpp:173
Done. 61/61 assertions succeeded in 648ms
Running test "buffered producer" @ /home/adamian/mbig/git/cppkafka/tests/producer_test.cpp:205
Done. 10/10 assertions succeeded in 1431ms
===============================================================================
All tests passed (103 assertions in 3 test cases)


Process finished with exit code 0

accelerated:
What about the watermark suggestion? Do you think it's worth implementing?

accelerated:
It's complaining about this line that it's assigned from uninitialized value...but that makes no sense. 
```
delivery_report_callback_ = config.get_delivery_report_callback();
```

accelerated:
Strange that a totally unrelated test fails

mfontanini:
Ignore that one. I saw that it fail once before, there's probably some race condition in the test itself.

accelerated:
I guess it should eventually be fixed as it may cause confusion for other PR's.

accelerated:
@mfontanini In regarding to your buffer comment above, would you accept this as a change to the Buffer constructor then?
```

    /**
     * Constructs a buffer from a pointer and a length
     *
     * \param data A pointer to some type
     * \param length The number of elements of type T
     */
    template <typename T>
    Buffer(const T* data, size_t length)
    : data_(reinterpret_cast<const DataType*>(data)), size_(length*sizeof(T)) {
    
    }

    /**
     * Constructs a buffer from a vector
     *
     * \param data The vector to be used as input
     */
    template <typename T>
    Buffer(const std::vector<T>& data)
    : Buffer(data.data(), data.size()) {

    }
   
   /**
     * Constructs a buffer from a raw array
     *
     * \param arr The raw array
     */
   template <typename array_type, size_t N>
   static Buffer make_buffer(array_type (&arr)[N]) {
          return Buffer(arr, N);
   }
```
This won't break any existing code as before you were mandating that `sizeof(T) == sizeof(DataType)` which means that anyone would just pass a byte array which also works in this new version. However this one is more intuitive.

mfontanini:
Based on that, then (unless I'm looking at this wrong):

```c++
// assume sizeof(int) == 4
vector<int> data = { 1, 2, 3 };
Buffer buffer1(data); // buffer1.size() == 3
Buffer buffer2(data.data(), data.size()); // buffer2.size() == 12
```

Why do you want to allow constructing buffers over arbitrary data sizes? There's things like endianness that would normally be an issue here. Also, how does this relate to the initial comment about `Buffer` you brought up? This doesn't fix your issue, does it?

accelerated:
Hi, well neither the old constructors nor the new constructors guarantee anything about endianness. Endian order types should be properly packed by the application prior to building the `Buffer` using some serialization scheme like BER/FlatBuffers/Avro/etc. All the Buffer class expects is a contiguous layout of types, nothing more. 
Regarding my initial comment, I think the Buffer constructors and other methods could be more user-friendly hence my proposal. 
Consider the following:
```
MyStruct myStructArray[] = {{},{},...};
vector<MyStruct> myStructVector{...};
vector<char> charVector{...}; //some byte buffer

// old constructor
Buffer b(reinterpret_cast<Buffer::DataType*>(myStructArray), sizeof(myStructArray));
Buffer b(reinterpret_cast<Buffer::DataType*>(myStructVector.data()), myStructVector.size()*sizeof(MyStruct));
Buffer b(charVector); // use the vector constructor since sizeof(T == Buffer::DataType == unsigned char)

// new constructor
Buffer b(myStructArray); 
Buffer b(myStructVector); 
Buffer b(charVector);
```
In the same token, allowing all copy/assignment/move operators from Buffer makes your life much easier in terms of coding. Because Buffer only contains a pointer and a size, even if you pass an rvalue and that goes away at the end of the call, the pointer to the data is non-owning so as long as that data is still valid then you're ok. I can understand restricting building a Buffer from a rvalue string object because that string object will free its internal buffer so the data pointer becomes invalid. But this is not the case with the Buffer class. Then:
```
MessageBuilder builder(message.get_topic());
const auto& key = message.get_key();
const auto& payload = message.get_payload();
builder.partition(message.get_partition())
       .key(Buffer(key.get_data(), key.get_size()))
       .payload(Buffer(payload.get_data(), payload.get_size()))
       .user_data(message.get_user_data());

// becomes 
MessageBuilder builder(message.get_topic());
builder.partition(message.get_partition())
       .key(message.get_key())
       .payload(message.get_payload())
       .user_data(message.get_user_data());
```
Also the `void produce(const MessageBuilder& builder)` can also have another overload `void produce(const Message& message)` so that producing logic inside `on_delivery_report` becomes:
```
if (should_produce) {
    MessageBuilder builder(message.get_topic());
    const auto& key = message.get_key();
    const auto& payload = message.get_payload();
    builder.partition(message.get_partition())
           .key(Buffer(key.get_data(), key.get_size()))
           .payload(Buffer(payload.get_data(), payload.get_size()))
           .user_data(message.get_user_data());
    if (message.get_timestamp()) {
        builder.timestamp(message.get_timestamp()->get_timestamp());
    }
    produce_message(builder);
    return;
}

// becomes
if (should_produce) {
    return produce_message(message);
}
```
Makes code more succinct, easier to grasp and less error prone.

mfontanini:
> Endian order types should be properly packed by the application prior to building the Buffer using some serialization scheme like BER/FlatBuffers/Avro/etc

Exactly. If you use e.g. protobuf, your code should look like this:

```c++
MyProtobufModel model = ...;
string buffer;
model.SerializeToString(&buffer);
MessageBuilder builder(message.get_topic());
builder.payload(buffer);
```

Not like this:

```c++
MyProtobufModel model = ...;
MessageBuilder builder(message.get_topic());
builder.payload(Buffer(&model, sizeof(model))); // oops, broken
```

If you want to send your data, either serialize it or if it's already serialized, use a proper pointer to it. Serializing anything into a message and using `reinterpret_cast` under the hood is a bad idea. Someone's going to misuse the class, end up sending a type that contains pointers (like `MyProtobufModel` above) and then complain that their code doesn't work. If you know what you're doing, just provide the proper pointer yourself.

Regarding copy constructors, that's that way for a reason. I guarantee that if copy/move constructors were added, within a few months there's going to be an issue brought up saying "I'm producing messages and I see that the key in them is garbage" and their code will look like this:

```c++
Buffer make_key(uint32_t v1, uint32_t v2) {
    const string result = to_string(v1) + ":" + to_string(v2);
    return Buffer(result); // oops
}

MessageBuilder builder("foo");
builder.key(make_key(a, b));
```

As I said, `Buffer` is a delicate class and I want to avoid potential issues as much as possible.

accelerated:
Ok so I guess you're trying to prevent people who don't know C++ from misusing the constructors, valid point. Maybe they should learn the hard way :). I always advocate code simplicity and proper documentation. 
Ok so for the `produce_method` would you consider overloading with a `Message` type as well? In case you already have a message, it would avoid rebuilding it in chunks with the MessageBuilder. I can add this to the current PR.

mfontanini:
Well, it's not just for people who don't know C++. It's better for the code to not let you shoot yourself in the foot than to allow silently doing so. Especially in cases like this where your code will likely not crash but instead you'll only realize when you start consuming on the other end of the topic and start seeing garbage, that's not a fun issue to have.

Regarding the overload of `produce` that sounds good. I wonder if it's better to handle this in `MessageBuilder` though as otherwise you'll need this overload on both `Producer` and `BufferedProducer`. Having a constructor like `MessageBuilder(const Message&)` that basically initializes the builder to whatever the message has and then lets you override specific attributes with whatever you want. This way you don't need to touch any of the producer classes.

accelerated:
> Especially in cases like this where your code will likely not crash but instead you'll only realize when you start consuming on the other end of the topic and start seeing garbage, that's not a fun issue to have.

Sounds just about the type of issues I'm having with librdkafka these days :))

> Regarding the overload of produce that sounds good. I wonder if it's better to handle this in MessageBuilder though as otherwise you'll need this overload on both Producer and BufferedProducer. Having a constructor like MessageBuilder(const Message&) that basically initializes the builder to whatever the message has and then lets you override specific attributes with whatever you want. This way you don't need to touch any of the producer classes.

I actually thought about that but personally I prefer the overload. I think it's awkward to build a message from a message using a builder class when you can just send it (if you have it). I see more the builder class as a way to piecemeal build a message. If you have one, it's unnecessary. Again I will leave the decision up to you. I have no issues with modifying the producer and the buffered producers.

mfontanini:
Fair enough, I agree that it's slightly odd to build a builder from a message. The overloads works.

mfontanini:
Thanks!

accelerated:
I was just about to say...the sooner you merge this the easier it will be to work on the other PR as there's some conflicts in the code. :)

mfontanini:
Yep, sorry about the delay. You'll probably have to do some rebasing.

PR #67, opened 2018-05-18
Pause/resume a consumer by topic

accelerated:
The base class `KafkaHandleBase` currently allows pausing/resuming by partitions which has fine granularity. In some use cases, the partitions are not really known or exposed to the application and furthermore, pausing/resuming action is mostly thought conceptually of at the topic level rather than at the partition level. These methods facilitate this.

Note: I debated whether or not these should go into the base class as well, but there are a few caveats with the producer. 
1) Pausing a producer can easily be done at the application level.
2) Producers can always produce to new topics (as long as they are on the registered broker list) ad-hoc and the partitioning is not available unless one queries the metadata from the brokers or uses a partitioning callback. Making this metadata call is fairly expensive and I'm not sure it's worth doing with each pause/resume. Metadata could be cached locally but this makes things more complicated. 
3) In the case where this is really needed, the base class method can be used.

### DIFF
--- a/include/cppkafka/topic_partition_list.h
+++ b/include/cppkafka/topic_partition_list.h
@@ -34,6 +34,7 @@
 #include <iosfwd>
 #include <algorithm>
 #include <vector>
+#include <set>
 #include <librdkafka/rdkafka.h>
 #include "macros.h"
 
@@ -55,14 +56,14 @@ CPPKAFKA_API TopicPartitionList convert(rd_kafka_topic_partition_list_t* topic_p
 CPPKAFKA_API TopicPartitionsListPtr make_handle(rd_kafka_topic_partition_list_t* handle);
 
 // Extracts a partition list subset belonging to the provided topics (case-insensitive)
-CPPKAFKA_API TopicPartitionList make_subset(const TopicPartitionList& partitions,
-                                            const std::vector<std::string>& topics);
+CPPKAFKA_API TopicPartitionList find_matches(const TopicPartitionList& partitions,
+                                             const std::set<std::string>& topics);
 
 // Extracts a partition list subset belonging to the provided partition ids
 // Note: this assumes that all topic partitions in the original list belong to the same topic
 //       otherwise the partition ids may not be unique
-CPPKAFKA_API TopicPartitionList make_subset(const TopicPartitionList& partitions,
-                                            const std::vector<int>& ids);
+CPPKAFKA_API TopicPartitionList find_matches(const TopicPartitionList& partitions,
+                                             const std::set<int>& ids);
 
 CPPKAFKA_API std::ostream& operator<<(std::ostream& output, const TopicPartitionList& rhs);
 
--- a/include/cppkafka/topic_partition_list.h
+++ b/include/cppkafka/topic_partition_list.h
@@ -54,6 +54,16 @@ CPPKAFKA_API TopicPartitionList convert(const TopicPartitionsListPtr& topic_part
 CPPKAFKA_API TopicPartitionList convert(rd_kafka_topic_partition_list_t* topic_partitions);
 CPPKAFKA_API TopicPartitionsListPtr make_handle(rd_kafka_topic_partition_list_t* handle);
 
+// Extracts a partition list subset belonging to the provided topics (case-insensitive)
+CPPKAFKA_API TopicPartitionList make_subset(const TopicPartitionList& partitions,
+                                            const std::vector<std::string>& topics);
+
+// Extracts a partition list subset belonging to the provided partition ids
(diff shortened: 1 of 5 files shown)


src/consumer.cpp
```diff
@@ -48,6 +52,23 @@ void Consumer::rebalance_proxy(rd_kafka_t*, rd_kafka_resp_err_t error,
     static_cast<Consumer*>(opaque)->handle_rebalance(error, list);
 }
 
+TopicPartitionList Consumer::get_matching_partitions(TopicPartitionList&& partitions,
+                                                     const vector<string>& topics) {
+    TopicPartitionList matches;
+    for (const auto& topic : topics) {
+        for (auto& partition : partitions) {
+            bool match = equal(topic.begin(), topic.end(), partition.get_topic().begin(),
```
mfontanini: Won't this blow up if `topic.size() > partition.get_topic().size()`?

src/consumer.cpp
```diff
@@ -48,6 +52,23 @@ void Consumer::rebalance_proxy(rd_kafka_t*, rd_kafka_resp_err_t error,
     static_cast<Consumer*>(opaque)->handle_rebalance(error, list);
 }
 
+TopicPartitionList Consumer::get_matching_partitions(TopicPartitionList&& partitions,
```
mfontanini: Why is this an rvalue reference and not a const ref?

src/consumer.cpp
```diff
@@ -125,6 +146,22 @@ void Consumer::unassign() {
     check_error(error);
 }
 
+void Consumer::pause() {
+    pause_partitions(get_assignment());
+}
+
+void Consumer::pause_topics(const std::vector<string>& topics) {
```
mfontanini: No `std` here please.

src/consumer.cpp
accelerated: Yeah looks like the signature was improved in C++17 with a fourth end iterator. I'll put a boundary check.

src/consumer.cpp
accelerated: I move it internally. I don't need to make a copy.

mfontanini: I didn't see the move. But anyway, in that case it should be taken by value, not by rvalue ref. The value will be moved into this function so won't be making any copies.

accelerated:
I made the subset selector function global since lots of methods take a topic partition list which could be used by topic instead. Therefore instead of providing overloads for all of them (very tedious) the selector by topic is now a util.

include/cppkafka/topic_partition_list.h
```diff
@@ -54,6 +54,16 @@ CPPKAFKA_API TopicPartitionList convert(const TopicPartitionsListPtr& topic_part
 CPPKAFKA_API TopicPartitionList convert(rd_kafka_topic_partition_list_t* topic_partitions);
 CPPKAFKA_API TopicPartitionsListPtr make_handle(rd_kafka_topic_partition_list_t* handle);
 
+// Extracts a partition list subset belonging to the provided topics (case-insensitive)
+CPPKAFKA_API TopicPartitionList make_subset(const TopicPartitionList& partitions,
```
mfontanini: This is okay here but I'm not sure about the name, "make_subset" doesn't really explain what the function does. Maybe `find_matches` or something like that? As much as this does return a subset of the input, the important thing is _how_ the subset is created and not the fact that it will return a subset.

accelerated:
Agreed, the loop flip did make things simpler. I also added the set instead of the vector and two test cases.

src/topic_partition_list.cpp
```diff
@@ -67,6 +70,40 @@ TopicPartitionsListPtr make_handle(rd_kafka_topic_partition_list_t* handle) {
...
+        }
+    }
+    return subset;
+}
+
+TopicPartitionList find_matches(const TopicPartitionList& partitions,
+                                const set<int>& ids) {
+    TopicPartitionList subset;
+    for (const auto& partition : partitions) {
+        for (const auto& id : ids) {
```
mfontanini: This whole loop can be replaced by 

```c++
if (ids.count(partition.get_partition()) > 0) {
    subset.emplace_back(partition);
}
```

I forgot you were doing the case insensitive comparisons for the topic case so you can't do this there, but this is mostly what I was going for with the set suggestion.

accelerated: I can put this code for the partition case. Not a bad idea. I think I've seen it elsewhere in your code

mfontanini:
Pushed some simplification, hopefully I didn't mess anything up as I edited it through the gh editor. I'll merge once build succeeds.

PR #68, opened 2018-05-19
Avoid SIGSEGV during destructing KafkaHandleBase (with set_log_callback())

azat:
Configuration should exist for logging, since it contains the copy of
the callback.

Stack traces:
```
  (gdb)
  #0  0x000055697b8f2140 in ?? ()
  #1  0x00007f10a393a514 in cppkafka::log_callback_proxy(rd_kafka_s const*, int, char const*, char const*) () from /usr/lib/libcppkafka.so.0.1
```
```
  (gdb) bt
  #0  0x00007f10a0d44a0d in __GI___pthread_timedjoin_ex (threadid=139697179305728, thread_return=0x7ffc7f574f70, abstime=0x0, block=<op
  timized out>) at pthread_join_common.c:89
  #1  0x00007f10a36d1640 in ?? () from /usr/lib/x86_64-linux-gnu/librdkafka.so.1
  #2  0x00007f10a36718ff in rd_kafka_destroy () from /usr/lib/x86_64-linux-gnu/librdkafka.so.1
  #3  0x00007f10a47f79af in __base_dtor  () at /usr/include/c++/7/bits/unique_ptr.h:268
  #4  __base_dtor  (this=0x55697b8ed260) at /usr/include/cppkafka/kafka_handle_base.h:63
```


include/cppkafka/kafka_handle_base.h
```diff
@@ -217,9 +217,9 @@ class CPPKAFKA_API KafkaHandleBase {
     std::vector<GroupInformation> fetch_consumer_groups(const char* name);
     void save_topic_config(const std::string& topic_name, TopicConfiguration config);
 
+    Configuration config_;
     HandlePtr handle_;
```
mfontanini: Good catch. Could you actually move the handle all the way down? That should be the first thing that gets destructed. I'm concerned about anything in `TopicConfigurationMap` having the same issue as the one you're describing.

azat: Yep, good suggestion. I force pushed the patch.

azat:
Are failed tests flacky?

```
/home/travis/build/mfontanini/cppkafka/tests/compacted_topic_processor_test.cpp:120: FAILED:
  CHECK( set_count == 2 )
with expansion:
  0 == 2
/home/travis/build/mfontanini/cppkafka/tests/compacted_topic_processor_test.cpp:121: FAILED:
  CHECK( delete_count == 1 )
with expansion:
  0 == 1
```

PR #73, opened 2018-05-24
Allow metadata object to be non-owning

accelerated:
Self explanatory. Also allows more consistent design to other similar handles

### DIFF
--- a/include/cppkafka/metadata.h
+++ b/include/cppkafka/metadata.h
@@ -140,7 +140,24 @@ private:
  */
 class CPPKAFKA_API Metadata {
 public:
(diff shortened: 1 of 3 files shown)


accelerated:
I want to do something similar to what you're doing with your message dr callback...where I create/own the metadata object and the just pass a reference during dr callback along with the message (I'm not passing at all the kafka handle, just its metadata for user-abstraction reasons). Creating the metadata is probably quite resource intensive as librdkafka has to gather all that info for all topics. So I'd rather do it once and then cache it until some "event occurs" like a reassigment. Anyway that's the jist of it. 
Also I may have a need for an empty metadata object.

PR #75, opened 2018-05-29
Added access to the internal Consumer in the backoff committer

accelerated:
Both `BufferedProducer` and `RoundRobinPollStrategy` allow access to the internal references so this class should too. It's also nice not having to hold on to the original Consumer reference if we ever need to modify something after the BackoffCommitter is constructed.

Also provided value constructor for the BackoffPerformer class.

### DIFF
--- a/include/cppkafka/utils/backoff_committer.h
+++ b/include/cppkafka/utils/backoff_committer.h
@@ -75,7 +75,7 @@ public:
     /**
      * \brief The error callback.
      * 
-     * Whenever an error occurs comitting an offset, this callback will be executed using
+     * Whenever an error occurs committing an offset, this callback will be executed using
      * the generated error. While the function returns true, then this is offset will be
      * committed again until it either succeeds or the function returns false.
      */
@@ -89,6 +89,18 @@ public:
      * \param consumer The consumer to use for committing offsets
      */
     BackoffCommitter(Consumer& consumer);
+    
+    /**
+     * \brief Constructs an instance using specified values
+     *
+     * \sa BackoffPerformer::BackoffPerformer
+     */
+    BackoffCommitter(Consumer& consumer,
+                     TimeUnit initial_backoff,
+                     TimeUnit backoff_step,
+                     TimeUnit maximum_backoff,
+                     BackoffPolicy policy,
+                     size_t maximum_retries);
 
     /**
      * \brief Sets the error callback
@@ -97,8 +109,7 @@ public:
      * \param callback The callback to be set
      */
     void set_error_callback(ErrorCallback callback);
-
-
+    
     /**
      * \brief Commits the given message synchronously
      *
@@ -118,6 +129,13 @@ public:
      * \param topic_partitions The topic/partition list to be committed
      */
     void commit(const TopicPartitionList& topic_partitions);
+    
+    /**
+     * \brief Get the internal Consumer object
+     *
+     * \return A reference to the Consumer
+     */
+    Consumer& get_consumer();
 private:
     // Return true to abort and false to continue committing
     template <typename T>
--- a/include/cppkafka/utils/backoff_committer.h
+++ b/include/cppkafka/utils/backoff_committer.h
@@ -89,18 +89,6 @@ public:
      * \param consumer The consumer to use for committing offsets
      */
     BackoffCommitter(Consumer& consumer);
-    
-    /**
-     * \brief Constructs an instance using specified values
-     *
-     * \sa BackoffPerformer::BackoffPerformer
-     */
-    BackoffCommitter(Consumer& consumer,
-                     TimeUnit initial_backoff,
-                     TimeUnit backoff_step,
-                     TimeUnit maximum_backoff,
-                     BackoffPolicy policy,
-                     size_t maximum_retries);
 
     /**
      * \brief Sets the error callback
--- a/include/cppkafka/utils/backoff_performer.h
+++ b/include/cppkafka/utils/backoff_performer.h
@@ -57,11 +57,26 @@ public:
     };
 
     /**
-     * Constructs an instance of backoff perform
+     * Constructs an instance of backoff performer
      * 
      * By default, the linear backoff policy is used
      */
     BackoffPerformer();
+    
+    /**
+     * \brief Constructs an instance of backoff performer
+     *
+     * \param initial_backoff The initial backoff interval
+     * \param backoff_step The backoff step interval
+     * \param maximum_backoff The maximum backoff interval
+     * \param policy The backoff policy
+     * \param maximum_retries The max number of retries
+     */
+    BackoffPerformer(TimeUnit initial_backoff,
+                     TimeUnit backoff_step,
+                     TimeUnit maximum_backoff,
+                     BackoffPolicy policy,
+                     size_t maximum_retries);
 
     /**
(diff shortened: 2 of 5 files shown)


include/cppkafka/utils/backoff_committer.h
```diff
@@ -88,6 +88,18 @@ class BackoffCommitter : public BackoffPerformer {
      * \param consumer The consumer to use for committing offsets
      */
     BackoffCommitter(Consumer& consumer);
+    
+    /**
+     * \brief Constructs an instance using specified values
+     *
+     * \sa BackoffPerformer::BackoffPerformer
+     */
+    BackoffCommitter(Consumer& consumer,
```
mfontanini: I don't like doing this at all. The day you want to add one more parameter, you're going to break everyone's code. I think I prefer having something like a `BackoffPerformerConfiguraiton` or something like that where each member is there. If you want to add something new, you'll break the ABI but at least people can just rebuild and that works. As it is, adding a new parameter breaks the API, unless you want to add another constructor but then it becomes unmaintainable.

accelerated: Ok fair enough. I was assuming that class was pretty much closed in terms of config. I'll remove the constructor.

mfontanini: I imagine it is but I've had this issue in the past with another library when I thought it was finished but then more arguments kept showing up and it ended up being quite messy.

accelerated:
Btw, I also wanted to add one helper method in `buffered_producer` named `sync_produce` (literally 2 lines) which would essentially be a combination of `produce` followed by `wait_for_acks`. The strawman argument :) is that...
1) it's easier to write one line instead of two each time I use it.
2) `flush` does the same under the hood (it could have well not called `wait_for_acks` and left that to the user instead)
3) there's `commit` and `async_commit` in the `Consumer` class so that's more for consistency purposes. 

If you're ok with this, I can roll it in this PR once I remove the above constructor, otherwise I can give you a separate PR. If you're not ok just let me know and I will skip it.

mfontanini:
That's not a strawman argument, that's valid :).That works! However, what if there's a produce failure? Wait for acks will exit when the failure is found, hence `sync_produce` will actually not produce anything. Are you planning on looping until it's _actually_ produced?

accelerated:
Hmmm not really, I was just planing on having it as a simple blocking function until the ack is received. Determining if there's an error can only be done via the supplied delivery failure callback. I could change the signature to `bool sync_produce()` but then I have to set some kind of internal flag in `on_delivery_report` to signal the error. That will only work if `sync_produce` is called from a single thread so not sure if it's worth it.

mfontanini:
Then I think the method doesn't exactly do what I'd expect. I mean, you're saying it "produces a message without buffering it and blocks until an ack is received" and that makes me think after that blocking method returns, I'll have produced the message but in reality that may not be the case. I don't know if I'd want to add such method given the semantics of it may be confusing, especially given the same behavior can be achieved in 2 lines of code.

accelerated:
Well the `sync` part has nothing to do with the production per-se. All it means is _blocking_. In the same token when you call `produce`, do you expect that it _produces_ something? We know that it can fail as well...
For flush it's a bit the same thing...why add `wait_for_acks` at the end when the user can simply do:
```
flush();
wait_for_acks();
```
Also 2 lines of code...

Actually I did want to incorporate a retry mechanism in both the regular produce and in the buffered/flush case (which is a requirement i have) but it's not very obvious to do unless you keep track of messages in some sort of map. I could probably implement it if you have some ideas, and if you're open to it / have suggestions?

I will remove the functions for now and we can add them later if we do the retry feature. For now I'll use the 2 liner. I will leave the Consumer access as that was my main change.

accelerated:
I think i found a super simple way of counting number of failed retries not using any external maps. Basically you replace the `user_data` pointer in the message with a `new tuple <void*, counter {some value}>` which wraps the old user data and the retry counter. In every `on_delivery` callback you get, if there's an error and you decide to re-enqueue it based on the user failure callback, you just decrement the counter and pass the wrapper again in the new builder. Same logic if it fails inside `flush` and you renqueue. When the counter reaches 0, you stop re-enqueuing.
Of course, when you raise any user callbacks, you pass tuple <0> inside the user data again so the aplication doesn't crash.
If you are fine with this method, i will give a separate PR after the remaining two are pulled.

accelerated:
Hi @mfontanini, I did the `sync_producer` (as you wished it would behave) + limiting the number of retries logic in this [branch here](https://github.com/accelerated/cppkafka/commit/acdaa0bc7b7ce21aff61e8313fc6f5314911f064#diff-d9488fa1d938111d1137c041302a170a). It's built on top of the two PRs you still have from me, so for now I won't issue a PR for this one. This functionality is quite important actually as it allows anyone to specify a retry limit w/o necessarily having to implement their own logic inside the ProduceFailureCallback. Since the functionality is pretty much universally needed, I would argue it's a very good thing to have it part of this framework.
Let me know what you think. You should review the changes in this order: Message, MessageBuilder, Producer, BufferedProducer. 
I've added a couple of internal functions in the public api, which can be avoided if you make the `BufferedProducer` a friend class, but then you may need to add other friends in the future so I'm not sure it's worth it - plus i'm not a fan of friend classes. Right now it's just documented that it's for _internal use only_.

accelerated:
Hi @mfontanini, here's what I propose: can you please accept the 2 outstanding PRs? I will then add the retry logic as a separate PR and we can wrap up the discussion there. I will adress *all* your comments in the PR and if it happens that you reject it, so be it, but i think the issue with producing from a dr_callback has a very easy fix and i will show you. But its easier to follow in a real PR page. The other issues i don't think are issues actually. Anyway...thanks.

include/cppkafka/utils/backoff_performer.h
```diff
@@ -57,11 +57,26 @@ class CPPKAFKA_API BackoffPerformer {
...
+    /**
+     * \brief Constructs an instance of backoff performer
+     *
+     * \param initial_backoff The initial backoff interval
+     * \param backoff_step The backoff step interval
+     * \param maximum_backoff The maximum backoff interval
+     * \param policy The backoff policy
+     * \param maximum_retries The max number of retries
+     */
+    BackoffPerformer(TimeUnit initial_backoff,
```
mfontanini: Looks like you only removed this constructor on the committer, this one's still here.

accelerated:
Ah ok, good catch, will update asap.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -535,6 +531,9 @@ void BufferedProducer<BufferType>::on_delivery_report(const Message& message) {
         // Increment the total successful transmissions
         ++total_messages_produced_;
     }
+    // Decrement the expected acks
+    --pending_acks_;
```
mfontanini: Do you mind explaining why this chunk was moved down here?

accelerated:
@mfontanini Ok I removed the last constructor. 

> Do you mind explaining why this chunk was moved down here?

Initially I decremented the ack counter at the top because the on_delivery_callback was throwing (originally when it was calling produce directly) hence I wanted to make sure that it gets decremented no matter what. Then as we removed the produce code, I moved it at the very bottom, because i want to make sure all user failure/success callbacks complete before any wait_for_acks return.

PR #79, opened 2018-06-05
Added purge (aka async_flush) functionality

accelerated:
Added asynchronous flush capability.

### DIFF
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -86,8 +86,8 @@ namespace cppkafka {
 template <typename BufferType>
 class CPPKAFKA_API BufferedProducer {
 public:
-    enum class EmptyBufferMethod { Flush,   ///< Empty the buffer and wait for acks from the broker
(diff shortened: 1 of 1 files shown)


accelerated:
If you can merge this that would be great! Cheers!

mfontanini:
I really thought I had merged this one, sorry :(. Anyway, I imagine you wanted the other one merged first so I went for that one.

accelerated:
Before I resolve the conflicts, I actually had one question for you...i was debating if `purge` was the right name here or perhaps `async_flush`. Do you have any preferences? I think `async_flush` is easier to understand what it does, but there there's `produce` and `sync_produce` so people might wonder why intermixing sync and async as it's not consistent.

mfontanini:
True, "purge" maybe sounds like you're getting rid of the messages rather than sending them. I'm not great at naming things but I think `async_flush` could work. I can't think of any other names... This is like some sort of "light flush" but that doesn't sound right.

accelerated:
ok cool, I'll change the name and push the update in a few mins.

mfontanini:
Thanks!

Issue #81, opened 2018-06-07
install incomplete


sachnk:
Hey --

The current master (`5cad740aea9fe0be5cef01c9959d8cb759c07c91`) doesn't seem to install properly. In particular, after doing `make install`, my application fails to compile due to:

```
/usr/local/include/cppkafka/consumer.h:41:10: fatal error: detail/callback_invoker.h: No such file or directory
```

On further inspection, it seems that the `detail` directory is not part of the CMake installation (`include/cppkafka/CMakeLists.txt`):

```
# Install headers including the auto-generated cppkafka.h
file(GLOB INCLUDE_FILES "*.h")
file(GLOB UTILS_INCLUDE_FILES "utils/*.h")
install(
    FILES ${INCLUDE_FILES}
    DESTINATION include/cppkafka
    COMPONENT Headers
)
install(
    FILES ${UTILS_INCLUDE_FILES}
    DESTINATION include/cppkafka/utils/
    COMPONENT Headers
)
```

This could be intentional, as the `detail` directory could be for private headers included by implementation. If that's the case, then publicly consumed header files shouldn't include anything from `detail`. Alternatively, we could simply add the `detail` directory to the CMake file installation recipe.

mfontanini:
Yeah, `detail` should actually be installed. It contains internal headers but those can be used inside other headers so they should be installed along. If you want to create a PR for this, that would be great :)

mfontanini:
Closed by #82

PR #82, opened 2018-06-07
#81: add detail headers to installation

sachnk:
Add missing cmake directives to install `detail` directory


mfontanini:
Thanks!

Issue #89, opened 2018-06-15
TopicPartitions don't contain errors (possible design change)


accelerated:
Many APIs from rdkafka which take a modifiable topic partition list, return errors on a per-partition basis inside the `rd_kafka_topic_partition_t` handle. Below is an example of two of them but there's others. cppkafka checks for errors on the API return code, but that always returns `RD_KAFKA_RESP_ERR_NO_ERROR` which is not useful. 

```
/**
 * @brief Pause producing or consumption for the provided list of partitions.
 *
 * Success or error is returned per-partition \p err in the \p partitions list.
 *
 * @returns RD_KAFKA_RESP_ERR_NO_ERROR
 */
RD_EXPORT rd_kafka_resp_err_t
rd_kafka_pause_partitions (rd_kafka_t *rk,
			   rd_kafka_topic_partition_list_t *partitions);



/**
 * @brief Resume producing consumption for the provided list of partitions.
 *
 * Success or error is returned per-partition \p err in the \p partitions list.
 *
 * @returns RD_KAFKA_RESP_ERR_NO_ERROR
 */
RD_EXPORT rd_kafka_resp_err_t
rd_kafka_resume_partitions (rd_kafka_t *rk,
			    rd_kafka_topic_partition_list_t *partitions);

/**
 * @brief Retrieve committed offsets for topics+partitions.
 *
 * The \p offset field of each requested partition will either be set to
 * stored offset or to RD_KAFKA_OFFSET_INVALID in case there was no stored
 * offset for that partition.
 *
 * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success in which case the
 *          \p offset or \p err field of each \p partitions' element is filled
 *          in with the stored offset, or a partition specific error.
 *          Else returns an error code.
 */
RD_EXPORT rd_kafka_resp_err_t
rd_kafka_committed (rd_kafka_t *rk,
		    rd_kafka_topic_partition_list_t *partitions,
		    int timeout_ms);
```

A suggested possible change would be to:
1) Enhance the `TopicPartition` object to also contain an `Error` object. This can be passed at constructions time by overloading the constructor. A setter method will also allow the error to be set at any time.
2) overload some of the cppkafka APIs which take a `const TopicPartitionList&` to also have a non-const overload. This new overload will modify the passed-in partition list by injecting the error(s), after the call to rdkafka API by using the setters in 1). The cppkafka API will now check _all_ partitions for errors before exiting.

An alternative (less invasive solution):
[...]

PR #90, opened 2018-06-18
added error check for partition list

accelerated:
Based on issue [89](https://github.com/mfontanini/cppkafka/issues/89)

### DIFF
--- a/include/cppkafka/kafka_handle_base.h
+++ b/include/cppkafka/kafka_handle_base.h
@@ -239,6 +239,8 @@ protected:
 
     void set_handle(rd_kafka_t* handle);
     void check_error(rd_kafka_resp_err_t error) const;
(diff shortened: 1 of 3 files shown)


accelerated:
Hi, yes I agree, if any partitions had an error, the API should have returned an error as well instead of success. The APIs I gave as example are not the only ones that have this behavior (you can grep rdkafka.h file for more but since you're not using all of them it does not matter here), so honestly I'm not sure what type of failures trigger it, maybe assigning invalid partitions, getting metadata for invalid partitions, setting bad offsets that are not within the special offset values, etc. I think that could be a question for librdkafka owner but he seems very busy at the moment.

accelerated:
@mfontanini is this ok?

mfontanini:
Yep, thanks!

PR #92, opened 2018-06-20
Fix for ref count on queue handles

accelerated:
Seems like the previous issue on queue ref count was finally [fixed](https://github.com/edenhill/librdkafka/issues/1792). I tested and everything seems fine now. Before, unless the handles were non-owning tests would fail.

### DIFF
--- a/src/consumer.cpp
+++ b/src/consumer.cpp
@@ -47,6 +47,17 @@ using std::equal;
 
 namespace cppkafka {
 
+// See: https://github.com/edenhill/librdkafka/issues/1792
+const int rd_kafka_queue_refcount_bug_version = 0x000b0500;
+Queue get_queue(rd_kafka_queue_t* handle) {
+    if (rd_kafka_version() <= rd_kafka_queue_refcount_bug_version) {
+        return Queue::make_non_owning(handle);
+    }
+    else {
+        return Queue(handle);
+    }
+}
+
 void Consumer::rebalance_proxy(rd_kafka_t*, rd_kafka_resp_err_t error,
                                rd_kafka_topic_partition_list_t *partitions, void *opaque) {
     TopicPartitionList list = convert(partitions);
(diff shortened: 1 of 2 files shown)


accelerated:
This is very weird, I recompiled twice and ran all tests and everything passes on my end.

mfontanini:
Two things:

* Travis checks out [a specific version](https://github.com/mfontanini/cppkafka/blob/master/.travis.yml#L10) of rdkafka and tests against it.
* The fact that this is fixed upstream doesn't mean the code can start using it. Anyone who uses a version of rdkafka that doesn't contain this fix will blow up. Even worse, that fix hasn't been released yet (it seems like it will be out for 0.11.5 but it's not out yet) so I don't think you can depend on it yet. Eventually you _could_ check the rdkafka version during compilation time and doing one thing or the other.

accelerated:
Ah, now I see...I was testing against master branch. Ok in that case we can just wait on this PR until 0.11.5 is ready. We can come back to this later.

accelerated:
Ok never mind my previous comment, I put in a compile-time check for the version. I'll remove it once you officially support 0.11.5

accelerated:
Just an FYI, based on the last [comment](https://github.com/edenhill/librdkafka/issues/1792), 0.11.5 should be due in 2 weeks approx.

accelerated:
@mfontanini any other concerns about this fix?

src/consumer.cpp
```diff
@@ -254,19 +254,33 @@ MessageList Consumer::poll_batch(size_t max_batch_size, milliseconds timeout) {
...
     Queue queue(Queue::make_non_owning(rd_kafka_queue_get_main(get_handle())));
+#else
+    Queue queue(rd_kafka_queue_get_main(get_handle()));
+#endif
     queue.disable_queue_forwarding();
     return queue;
 }
 
 Queue Consumer::get_consumer_queue() const {
+#if RD_KAFKA_VERSION <= 0x000b0500
```
mfontanini: I think this should do `if (rd_kafka_version() <= ...)`. There's an issue here if you build the library on a system that has rdkafka >= 0.11.5 but you deploy it on one that has rdkafka < 0.11.5. In that case, this won't really work.

mfontanini: Also I'd probably create a function within this cpp file that gets an rdkafka queue pointer and returns a `Queue` so you don't have the exact same check deplicated N times.

accelerated: Ok, runtime check works too, you're right, it's safer. I'll wrap this up as per your comments.

accelerated:
Done!

accelerated:
Thanks for approving! Can you please merge? Cheers!!

mfontanini:
I didn't merge this because I don't want to merge something that depends on an unreleased version of rdkafka, especially when under the hood this should still work the old way. If they do release it in 10 days, I'll merge it by then.

accelerated:
But the version I'm comparing against is already [released]!(https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L151)! 

The one 0.11.5 which you are referring to will be `0x000b05ff`

mfontanini:
The [fix](https://github.com/edenhill/librdkafka/commit/26418ae7351e7a7c7f54aac97addcafcbdaebadd) for this issue is not released yet. It's tagged as some pre release for version v0.11.5 but it's not out yet. I don't see a point in having version checks for yet unreleased versions. What if for some reason this fix doesn't get released on that version?

accelerated:
ok let's wait then, there's no rush anyway. Seems like it's very imminent.

accelerated:
I think now it should be [safe](https://github.com/edenhill/librdkafka/issues/1792) based on last comments. Also the version was bumped to final 0.11.5

mfontanini:
Sweet, thanks!

PR #93, opened 2018-06-20
Invoke error callback if present instead of log callback

accelerated:
Small change in case Consumer destructor throws. If the user has an error callback registered, invoke it instead of the log callback since we know the rdkafka error enum, otherwise fall back unto the log callback. `close()` throws a `HandleException` so we catch that instead.


mfontanini:
Thanks!

PR #94, opened 2018-06-22
Specific linking option for rdkafka library

accelerated:
Gives an option at build time to statically or dynamically link the rdkafka library. The issue is that if both .a and .so files are present in the same directory (or .lib and .dll for windows), cmake seems to prefer the shared library by default. The only way to control this is to force `find_path` to look for more specific names first and then the more generic names (without a suffix) afterward.
If `.a .so .lib and .dll` files are not there, the default cmake search pattern will be used.


accelerated:
@mfontanini Hi, would it be possible to review the PRs I submitted? They're quite straightforward. The main ones are this one and the pause/resume. The others can wait but ideally if you can wrap them up it would be nice. Thanks!

mfontanini:
Looks good!

PR #98, opened 2018-07-02
Fix #97: `struct` switched to `class`.

### LINKED ISSUE #97: MSVC name-mangling issue
```
LNK2019	unresolved external symbol "public: class std::shared_ptr<struct cppkafka::Internal> __thiscall cppkafka::MessageInternal::get_internal(void)const "
(?get_internal@MessageInternal@cppkafka@@QBE?AV?$shared_ptr@UInternal@cppkafka@@@std@@XZ)
referenced in function "private: class cppkafka::Message & __thiscall cppkafka::Message::load_internal(void)"
 (?load_internal@Message@cppkafka@@AAEAAV12@XZ)	.....\cppkafka.lib(message.obj)	1	
```
MSVC trying to find `?get_internal@MessageInternal@cppkafka@@QBE?AV?$shared_ptr@UInternal@cppkafka@@@std@@XZ`
but in `message_internal.obj` it mangled in `?get_internal@MessageInternal@cppkafka@@QBE?AV?$shared_ptr@VInternal@cppkafka@@@std@@XZ`

The difference: `@UInternal@` vs `@VInternal@` - `struct` vs `class` (accorging to https://en.wikiversity.org/wiki/Visual_C%2B%2B_name_mangling and `undname` tool).

for private use `Internal` is declared as a class:
https://github.com/mfontanini/cppkafka/blob/master/include/cppkafka/message_internal.h#L39

for public use it is a struct:
https://github.com/mfontanini/cppkafka/blob/master/include/cppkafka/message.h#L46

I propose to change public declaration to class too. Can I make a pull request?

dmpas:
To avoid name mangling difference in MSVC.

I'm not sure this class should be marked as API.

### DIFF
--- a/include/cppkafka/message.h
+++ b/include/cppkafka/message.h
@@ -43,7 +43,7 @@
 namespace cppkafka {
 
 class MessageTimestamp;
-struct Internal;
+class Internal;
 
 /**
  * \brief Thin wrapper over a rdkafka message handle


mfontanini:
I think what you said in the ticket makes sense. This class should probably live inside the `detail` directory (and probably namespace) but that's another issue. Thanks!

PR #99, opened 2018-07-03
Fix #83: define LIBRDKAFKA_STATICLIB

### LINKED ISSUE #83: Unresolved external when linking into DLL on Windows
Sucessfully built both rdkafka and cppkafka. Both *.lib are specified in linker output

```
LNK2019	unresolved external symbol __imp__rd_kafka_conf_new referenced in function 
"public: __thiscall cppkafka::Configuration::Configuration(void)" (??0Configuration@cppkafka@@QAE@XZ) 
d:\project-path\cppkafka.lib(configuration.obj)
```

and many similar errors

It seems, that linker can't find **librdkafka** functions, which are used in **cppkafka**
Any ideas why this happens?

image

dmpas:
https://github.com/edenhill/librdkafka/blob/1827da4510dd700d3a1e1c01b94c123d3528199e/src/rdkafka.h#L69
```c
#ifdef LIBRDKAFKA_STATICLIB
#define RD_EXPORT
#else
#ifdef LIBRDKAFKA_EXPORTS
#define RD_EXPORT __declspec(dllexport)
#else
#define RD_EXPORT __declspec(dllimport)
#endif
```

if `LIBRDKAFKA_STATICLIB` is not defined, `RD_EXPORT` is defined as dllimport and MSVC trying to find __imp-entries.
So, we must define `LIBRDKAFKA_STATICLIB` in case of linking static rdkafka library.


cmake/FindRdKafka.cmake
```diff
@@ -1,6 +1,7 @@
 # Override default CMAKE_FIND_LIBRARY_SUFFIXES
 if (CPPKAFKA_RDKAFKA_STATIC_LIB)
     if (MSVC)
+        add_definitions("-DLIBRDKAFKA_STATICLIB")
```
mfontanini: I don't think this is the best place to put this. Can you move this into the main CMakeLists.txt file? This is supposed to find rdkafka, I wouldn't expect calling "find rd kafka" to add a definition under the hood.

dmpas: Ok.

PR #107, opened 2018-07-25
Fix #104: memory leak in poll_batch

### LINKED ISSUE #104: Memory leak when cosuming with poll_batch function
Hey. I'm trying to implement a consumer using cppkafka. The example code which uses `Message Consumer::poll(milliseconds timeout)` function works perfectly. However, when using `MessageList Consumer::poll_batch(size_t max_batch_size, milliseconds timeout)` function there occurs a memory leak.

Below is the consumer code:
```
#include <chrono>
#include <iostream>
using namespace std::chrono_literals;

int main() {
  cppkafka::Consumer consumer_(cppkafka::Configuration(
    {{"metadata.broker.list", "MY_BROKER_LIST"},
     {"group.id", "MY_GROUP_ID"},
     {"enable.auto.commit", false},
     {"client.id", "MY_CLIENT_ID"},
     {"auto.offset.reset", "latest"},
     {"api.version.request", true}}));
  consumer_.subscribe({"MY_TOPIC"});
  for (;;) {
    for (cppkafka::Message& msg: consumer_.poll_batch(1024, 0ms)) {
      if (!msg) {
        continue;
      }
      std::cout << "Got message ...";
    }
  }
}
```
Below is the result after runing the executable with valgrind for 10 second.
```
LEAK SUMMARY:
   definitely lost: 2,616,174 bytes in 109,002 blocks
   indirectly lost: 0 bytes in 0 blocks
     possibly lost: 2,936 bytes in 29 blocks
   still reachable: 50,043,479 bytes in 124,985 blocks
        suppressed: 0 bytes in 0 blocks
```

Env:
- Ubuntu 18.04
- cppkafka: latest master (d6f812920712c79926fb0871f811543978425905)
- librdkafka: latest master (44242a464c43e09c685f47a7f3dca2963b10e2a9)

shashank88:
poll_batch currently leaks memory while initialising the queue
returned by rd_kafka_queue_get_consumer. The fix as suggested
by @mfontanini as done here is to initialise the queue with a
Queue so it's cleaned up when going out of scope.

### DIFF
--- a/src/consumer.cpp
+++ b/src/consumer.cpp
@@ -261,8 +261,9 @@ MessageList Consumer::poll_batch(size_t max_batch_size) {
 
 MessageList Consumer::poll_batch(size_t max_batch_size, milliseconds timeout) {
     vector<rd_kafka_message_t*> raw_messages(max_batch_size);
-    rd_kafka_queue_t* queue = rd_kafka_queue_get_consumer(get_handle());
-    ssize_t result = rd_kafka_consume_batch_queue(queue, timeout.count(), raw_messages.data(),
+    // Note that this will leak the queue when using rdkafka < 0.11.5 (see get_queue comment)
+    Queue queue(get_queue(rd_kafka_queue_get_consumer(get_handle())));
+    ssize_t result = rd_kafka_consume_batch_queue(queue.get_handle() , timeout.count(), raw_messages.data(),
                                                   raw_messages.size());
     if (result == -1) {
         check_error(rd_kafka_last_error());


src/consumer.cpp
```diff
@@ -261,8 +261,8 @@ MessageList Consumer::poll_batch(size_t max_batch_size) {
 
 MessageList Consumer::poll_batch(size_t max_batch_size, milliseconds timeout) {
     vector<rd_kafka_message_t*> raw_messages(max_batch_size);
-    rd_kafka_queue_t* queue = rd_kafka_queue_get_consumer(get_handle());
-    ssize_t result = rd_kafka_consume_batch_queue(queue, timeout.count(), raw_messages.data(),
+    Queue queue(get_queue(rd_kafka_queue_get_consumer(get_handle())));
```
mfontanini: This won't actually fix the issue if someone's using rdkafka version < 0.11.5 (see the comment on `get_queue`). Maybe it's worth adding a comment saying that right above this line? e.g. "Note that this will leak the queue when using rdkafka < 0.11.5 (see get_queue comment)`.

shashank88: Squashed the comment to the same commit: 
https://github.com/mfontanini/cppkafka/pull/107

mfontanini:
Thanks!

PR #112, opened 2018-08-30
Fix #111: Rename LogLevel enums

### LINKED ISSUE #111: LogLevel enum clashes with syslog.h
The `LogLevel` enum in `cppkafka/logging.h` doesn't play nicely when `syslog.h` is included prior to its definition (not surprisingly, based on the comment preceding it).

```
$ grep "LOG_EMERG" /usr/include/sys/syslog.h
#define LOG_EMERG       0       /* system is unusable */
```

From `cppkafka/logging.h`:
```cpp
 // Based on syslog.h levels
 enum class LogLevel : int {
    LOG_EMERG   = 0,    /* system is unusable */
```

zenonparker:
As described in Issue #111

### DIFF
--- a/include/cppkafka/detail/callback_invoker.h
+++ b/include/cppkafka/detail/callback_invoker.h
@@ -101,7 +101,7 @@ public:
                 try {
                     // Log it
                     handle_->get_configuration().get_log_callback()(*handle_,
-                                                                    static_cast<int>(LogLevel::LOG_ERR),
+                                                                    static_cast<int>(LogLevel::LogErr),
                                                                     library_name,
                                                                     error_msg.str());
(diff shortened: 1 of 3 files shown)


mfontanini:
Thanks for this. This will break people's code but it won't work as it is so we have no choice. @accelerated heads up, not sure if you're using this outside of cppkafka.

accelerated:
Well the alternative non-breaking solution would have been to #undef LOG_EMERG & others right above the LogLevel enum.

mfontanini:
Right but then if you have an application that uses both syslog and cppkafka, then you'll have an issue when trying to use those macros.

PR #118, opened 2018-10-10
Allocators

accelerated:
Support for allocators for batch consuming and for buffered producer.
Changed MessageList back to `std::vector<Message>` for consistency with the Allocator API, however the `MessageList typedef` is still there for backward compatibility.


accelerated:
Maybe you could merge this one first? So when you test the headers, you'll also implicitly test this?

cfinkenstadt:
👍

PR #128, opened 2018-10-25
Added time_point overloads for creating timestamps.

### LINKED ISSUE #15: relative time as timestamp
The message timestamp is expressed as a relative time duration (``std::chrono::milliseconds``) [here](https://github.com/mfontanini/cppkafka/blob/master/include/cppkafka/message.h#L195).

That seems like the wrong type, it really should be a ``time_point``, right?

accelerated:
Changes to `BasicMessageBuilder` and `MessageTimestamp` classes to allow `std::chrono::time_point` overloads.

**Note** that `MessageTimestamp` is created internally by the `Message` object, however since it's constructor was public, I figured others could use this class to store timestamps outside of the `Message` implementation.

Fixes #15

### DIFF
--- a/include/cppkafka/message.h
+++ b/include/cppkafka/message.h
@@ -240,12 +240,19 @@ public:
     };
 
     /**
-     * Constructs a timestamp object
+     * Constructs a timestamp object using a 'duration'.
      */
     MessageTimestamp(std::chrono::milliseconds timestamp, TimestampType type);
+    
+    /**
+     * Constructs a timestamp object using a 'time_point'.
+     */
+    template <typename Clock, typename Duration = typename Clock::duration>
+    MessageTimestamp(std::chrono::time_point<Clock, Duration> timestamp, TimestampType type);
 
     /**
-     * Gets the timestamp value
+     * Gets the timestamp value. If the timestamp was created with a 'time_point',
+     * the duration represents the number of milliseconds since epoch.
      */
     std::chrono::milliseconds get_timestamp() const;
 
--- a/tests/producer_test.cpp
+++ b/tests/producer_test.cpp
@@ -164,7 +164,7 @@ TEST_CASE("simple production", "[producer]") {
     SECTION("message with key") {
         const string payload = "Hello world! 2";
         const string key = "such key";
-        const milliseconds timestamp{15};
+        auto timestamp = std::chrono::system_clock::now();
         Producer producer(config);
         producer.produce(MessageBuilder(KAFKA_TOPICS[0]).partition(partition)
                                                      .key(key)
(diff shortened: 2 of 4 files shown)


accelerated:
**Note2** we could get rid of the template parameters and just force  the time point to be `std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds>`. Not sure if it's a good idea though.

mfontanini:
Yeah, I was using steady clock as an example of a clock that makes no sense out of the computer it was generated on, but any other will do.

tests/producer_test.cpp
```diff
@@ -181,7 +183,7 @@ TEST_CASE("simple production", "[producer]") {
         CHECK(message.get_partition() == partition);
         CHECK(!!message.get_error() == false);
         REQUIRE(!!message.get_timestamp() == true);
-        CHECK(message.get_timestamp()->get_timestamp() == timestamp);
+        CHECK(message.get_timestamp()->get_timestamp() == duration_cast<milliseconds>(timestamp.time_since_epoch()));
```
arvidn: this test does not exercise the new constructor. If it did, you would notice that the definition is not available, and will produce a link error

include/cppkafka/message.h
```diff
@@ -240,12 +240,19 @@ class CPPKAFKA_API MessageTimestamp {
...
-     * Constructs a timestamp object
+     * Constructs a timestamp object using a 'duration'.
      */
     MessageTimestamp(std::chrono::milliseconds timestamp, TimestampType type);
+    
+    /**
+     * Constructs a timestamp object using a 'time_point'.
+     */
+    template <typename Clock, typename Duration = typename Clock::duration>
+    MessageTimestamp(std::chrono::time_point<Clock, Duration> timestamp, TimestampType type);
```
arvidn: it's not clear to me why it makes sense to have this template constructor be public, since its definition is only available in the .cpp file, and there's no explicit instantiation of it. In fact, I don't see *any* instantiations of this constructor

mfontanini: Yeah, the definition should be on the header. It would never work the way it is.

mfontanini: Actually, true, why is this constructor even here? Nothing uses it and nothing can construct a timestamp besides cppkafka itself.

mfontanini: I put up https://github.com/mfontanini/cppkafka/pull/129 to fix this. I don't think we need that overload at all

accelerated:
@mfontanini see my note in the first comment I made.

include/cppkafka/message.h
accelerated: I can move the def from the .cpp file. I initially did it w/o templates, and then after you said it was ok to templetize based on the clock type I just added the templates but forgot to move it to the header file. My bad.

PR #129, opened 2018-10-26
Remove MessageTimestamp constructor from time_point

mfontanini:
Relates to #128

### DIFF
--- a/include/cppkafka/message.h
+++ b/include/cppkafka/message.h
@@ -245,12 +245,6 @@ public:
     MessageTimestamp(std::chrono::milliseconds timestamp, TimestampType type);
     
     /**
-     * Constructs a timestamp object using a 'time_point'.
-     */
-    template <typename Clock, typename Duration = typename Clock::duration>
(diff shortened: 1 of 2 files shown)


mfontanini:
@accelerated can you look into this one? I don't think we need that constructor, given the only one that constructs `MessageTimestamp`s is cppkafka itself.

accelerated:
See my comment in the Issue thread. I can move it to the header file, that was my oversight or remove it, but I did make a note about that when I submitted the PR.

mfontanini:
I don't think a user should be able to create `MessageTimestamp` (right now it can't). The user shouldn't be able to specify a `MessageTimestamp::TimestampType` because when calling `rd_kafka_producev` that goes away: at that point it's just a millisecond count.

Given that nothing constructs `MessageTimestamp` besides the `Message`, there's no point in allowing creating them from `time_point`

mfontanini:
Yep, those are good ideas.

PR #133, opened 2018-11-06
Message timestamp refactoring and log level changes

accelerated:
Small PR including:
* Refactoring the `MessageTimestamp` into its own separate file as well as making its constructor private.
* Added the `KafkaHandleBase::set_log_level()` method.
* Set default logging level to _Error_ in both `Producer` and `Consumer` class. 

_Note that the `Producer` class was setting the level to `DEBUG` in the constructor, whereas the Consumer was not setting anything. I'm not sure if this was intentional or just an oversight. Also the `Producer` was using a hard-coded log level value...perhaps a leftover from a debugging session._

### DIFF
--- a/include/cppkafka/kafka_handle_base.h
+++ b/include/cppkafka/kafka_handle_base.h
@@ -110,7 +110,7 @@ public:
     void set_timeout(std::chrono::milliseconds timeout);
     
     /**
-     * @brief Sets the log level
+     * \brief Sets the log level
      */
      void set_log_level(LogLevel level);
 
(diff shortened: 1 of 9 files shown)


include/cppkafka/kafka_handle_base.h
```diff
@@ -107,6 +108,11 @@ class CPPKAFKA_API KafkaHandleBase {
      * \param timeout The timeout to be set
      */
     void set_timeout(std::chrono::milliseconds timeout);
+    
+    /**
+     * @brief Sets the log level
```
mfontanini: Minor: use `\` rather than `@` here.

accelerated:
>Looks good, just a minor comment. I'm not sure when that log level was set in the producer, but it's possible it was while debugging.

That's what I thought. In any case Error level is fine as default.

mfontanini:
Thanks!

PR #134, opened 2018-11-12
Added support for message status and setting the event mask

accelerated:
Added support for `rd_kafka_conf_set_events` and `rd_kafka_message_status`


mfontanini:
Thanks!

PR #136, opened 2018-11-13
Remove setting log level in the constructor

accelerated:
Setting the log level in the producer and consumer constructor can interfere with the user setting the `log_level` rdkafka option which essentially gets overriden to Err level. It's best to either let the user specify it via `log_level` rdkafka option or via the `KafkaHandleBase::set_log_level` api. 

As an alternative, we could check if the user has set the `log_level` option and if so, not call the `set_log_level` api. But it feels hackish.


mfontanini:
This makes sense, the log level setting shouldn't have been there from the start. Thanks!

PR #138, opened 2018-11-17
Add support for constructing Buffer from std::array

tgaldes:


include/cppkafka/buffer.h
```diff
@@ -92,6 +92,17 @@ class CPPKAFKA_API Buffer {
...
     }
 
+    /**
+     * Constructs a buffer from an array
+     *
+     * \param data The the array to be used as input
+     */
+    template <typename T, size_t N>
+    Buffer(const std::array<T, N>& data)
+    : data_(reinterpret_cast<const DataType*>(&data[0])), size_(N * sizeof(T)) {
```
mfontanini: Minor: can you use `data.data()` and `data.size()` rather than `&data[0]` and `N * sizeof(T)` here?

accelerated:
Maybe we can also have an overload for raw arrays `Buffer(T(&arr)[N])` ?

mfontanini:
Yeah, that's a good idea!

tgaldes:
@accelerated is there a difference between construction with the raw array and the constructor Buffer(const T*, size_t size) here: 
https://github.com/mfontanini/cppkafka/blob/master/include/cppkafka/buffer.h#L76

None of the code in message_builder.h uses that template from line 76, so maybe the solution is to overload payload() of BasicMessageBuilder with payload(const T* value, size_t size), and also overload the implicit conversion between ConcreteMB and MBuilder with something that will construct the buffer of the MBuilder with that same pointer and size to that no data needs to be copied.

include/cppkafka/buffer.h
tgaldes: will do! first time working with std::arrays

accelerated:
@tgaldes, the difference is simplicity and easy of use. Just like you can skip the `std::vector` and `std::array` constructors because eventually everything is convertible to a T* and a size, but they're nice to have. `BasicMessageBuilder` takes `BufferType` in all its methods and those can be a `Buffer`, so no need to overload with `(const T*, size_t)`. It would make the message builder class unnecessarily complicated.

include/cppkafka/buffer.h
mfontanini: Perfect, that works! One more thing: can you add a deleted constructor taking a `std::array<T, N>&&` just like there is for `vector` and `string`? e.g. [this one](https://github.com/mfontanini/cppkafka/blob/master/include/cppkafka/buffer.h#L97). This will make sure people can't misuse `Buffer` and construct it from a temporary `array`, which will end up making the buffer contain garbage once it goes out of scope.

tgaldes: should be fixed

mfontanini:
Thanks! The PR build is flapping but it's unrelated.

tgaldes:
@accelerated Here's the addition you requested: https://github.com/mfontanini/cppkafka/pull/140

PR #144, opened 2018-11-29
Header fixes and header copy considerations

accelerated:
### Scope of fixes
* `ClonerPtr` was not copying the cloner function on copy. Also few other helper functions were added to this class.
* `MessageBuilder` class was missing header support in a few functions which caused the feature not to work with the `BufferedProducer` class. 
* Added `BasicMessageBuilder::headers` overload to allow taking header lists as well.
* Removed asserts in the `HeaderList::iterator` access so as to allow iterating even if the list is empty. This makes the code easier. Same for `size()` and `empty()` methods which should work even if there's no header handle.
* Added set header methods in the `Message` class to support `rd_kafka_message_set_headers` API which is useful if re-producing a `Message` with different headers. (see `BasicMessageBuilder::BasicMessageBuilder(const Message&)` or `BufferedProducer::send(const Message&)` APIs)

### Current issues -- too many header copies!! 
While debugging the `BufferedProducer` I realized that the headers (if present) are being copied too many times. For instance when calling `BufferedProducer::produce(const MessageBuilder&)`:
* I create a `ConcreteMessageBuilder<string>` -> creation of headers. I can't use directly a `MessageBuilder` for architecture considerations...
* Call produce() which calls the `MessageBuilder` conversion constructor. This causes the headers to be copied once (part of the fix above).
* Inside `produce`, assuming we have internal data (optional), we clone the `MessageBuilder` class. 2nd copy. (part of the fix above)
* Then `BufferedProducer::produce_message` calls the `Producer::produce(const MessageBuilder&)` (instead of the move produce) which causes a 3rd copy. At this point the headers are moved into rdkafka which takes ownership of them. 

If the headers are small, this is probably not too much of an issue, but if the headers are large buffers, and there's multiple headers, then it becomes a bottleneck. I'm not sure how to solve this, but one idea would be to make the `MessageBuilder` class non-header-owning. This makes sense conceptually because `MessageBuilder` members are `Buffers` which by themselves are non-owning, so why would the headers be copied? This would avoid the 1st and 2nd copy and inside `Producer::produce` we can make a hard copy (i.e. calling `rd_kafka_headers_copy(MessageBuilder::header_list()::get_handle())` since the copy constructor would not work in this case as it's non-owning. 

One problem with making the `MessageBuilder` non-owning is that when calling `MessageBuilder::header` for adding an individual header, we will create underneath a header list but that list will be non-owning. This is a bit counter-intuitive. Also when calling `Producer::produce`, this will make another copy, but then who will delete the handle non-owned by the `MessageBuilder`? 
If we remove the copy inside `Producer::produce`, then if someone is building a `MessageBuilder` from a `ConcreteMessageBuilder`, that concrete class will have to `release` its internal ownership as it will be passed unto rdkafka. This is an extra manual step which will be ugly. 

I have fixed the issues but I have **not** made the proposed enhancements to the `MessageBuilder` class as I wasn't sure if you had other solutions.

### DIFF
--- a/include/cppkafka/buffer.h
+++ b/include/cppkafka/buffer.h
@@ -81,6 +81,11 @@ public:
             throw Exception("Invalid buffer configuration");
         }
     }
+    
+    template <typename Iter>
+    Buffer(const Iter first, Iter last)
+    : Buffer(&*first, std::distance(first, last)) {
+    }
 
     /**
      * Constructs a buffer from a vector
--- a/include/cppkafka/buffer.h
+++ b/include/cppkafka/buffer.h
@@ -82,6 +82,12 @@ public:
         }
     }
     
+    /**
+     * Constructs a buffer from two iterators in the range [first,last)
+     *
+     * \param first An iterator to the start of data
+     * \param last An iterator to the end of data (not included)
+     */
     template <typename Iter>
     Buffer(const Iter first, Iter last)
     : Buffer(&*first, std::distance(first, last)) {
--- a/include/cppkafka/buffer.h
+++ b/include/cppkafka/buffer.h
@@ -93,7 +93,9 @@ public:
         static_assert(sizeof(T) == sizeof(DataType), "sizeof(T) != sizeof(DataType)");
     }
 
-    // Don't allow construction from temporary vectors
+    /**
+     * Don't allow construction from temporary vectors
+     */
     template <typename T>
     Buffer(std::vector<T>&& data) = delete;
 
@@ -108,7 +110,9 @@ public:
         static_assert(sizeof(T) == sizeof(DataType), "sizeof(T) != sizeof(DataType)");
     }
 
-    // Don't allow construction from temporary arrays
+    /**
+     * Don't allow construction from temporary arrays
+     */
     template <typename T, size_t N>
     Buffer(std::array<T, N>&& data) = delete;
 
@@ -120,9 +124,11 @@ public:
      */
     Buffer(const std::string& data); 
 
-    // Don't allow construction from temporary strings
+    /**
+     * Don't allow construction from temporary strings
+     */
     Buffer(std::string&&) = delete;
-
+    
     Buffer(const Buffer&) = delete;
     Buffer(Buffer&&) = default;
     Buffer& operator=(const Buffer&) = delete;
--- a/include/cppkafka/buffer.h
+++ b/include/cppkafka/buffer.h
@@ -89,7 +89,7 @@ public:
      * \param last An iterator to the end of data (not included)
      */
     template <typename Iter>
-    Buffer(const Iter first, Iter last)
+    Buffer(const Iter first, const Iter last)
     : Buffer(&*first, std::distance(first, last)) {
     }
 
--- a/include/cppkafka/clonable_ptr.h
+++ b/include/cppkafka/clonable_ptr.h
@@ -60,8 +60,7 @@ public:
      * \param rhs The pointer to be copied
      */
     ClonablePtr(const ClonablePtr& rhs)
-    : handle_(rhs.get_cloner() ? std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter()) :
-                                 std::unique_ptr<T, Deleter>(rhs.get(), rhs.get_deleter())),
+    : handle_(std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter())),
       cloner_(rhs.get_cloner()) {
 
     }
@@ -73,8 +72,7 @@ public:
      */
     ClonablePtr& operator=(const ClonablePtr& rhs) {
         if (this != &rhs) {
-            handle_ = rhs.get_cloner() ? std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter()) :
-                                         std::unique_ptr<T, Deleter>(rhs.get(), rhs.get_deleter());
+            handle_ = std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter());
             cloner_ = rhs.get_cloner();
         }
         return *this;
@@ -130,7 +128,7 @@ private:
      * \brief Clones the internal pointer using the specified cloner function.
      */
     T* clone() const {
-        return cloner_ ? cloner_(handle_.get()) : handle_.get();
+        return cloner_ ? cloner_(get()) : get();
     }
     
     std::unique_ptr<T, Deleter> handle_;
--- a/include/cppkafka/clonable_ptr.h
+++ b/include/cppkafka/clonable_ptr.h
@@ -60,9 +60,9 @@ public:
      * \param rhs The pointer to be copied
      */
     ClonablePtr(const ClonablePtr& rhs)
-    : handle_(rhs.cloner_ ? std::unique_ptr<T, Deleter>(rhs.cloner_(rhs.handle_.get()), rhs.handle_.get_deleter()) :
-                            std::unique_ptr<T, Deleter>(rhs.handle_.get(), rhs.handle_.get_deleter())),
-      cloner_(rhs.cloner_) {
+    : handle_(rhs.get_cloner() ? std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter()) :
+                                 std::unique_ptr<T, Deleter>(rhs.get(), rhs.get_deleter())),
+      cloner_(rhs.get_cloner()) {
 
     }
 
@@ -72,11 +72,11 @@ public:
      * \param rhs The pointer to be copied
      */
     ClonablePtr& operator=(const ClonablePtr& rhs) {
-        if (this == &rhs) {
-            return *this;
+        if (this != &rhs) {
+            handle_ = rhs.get_cloner() ? std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter()) :
+                                         std::unique_ptr<T, Deleter>(rhs.get(), rhs.get_deleter());
+            cloner_ = rhs.get_cloner();
         }
-        handle_ = rhs.cloner_ ? std::unique_ptr<T, Deleter>(rhs.cloner_(rhs.handle_.get()), rhs.handle_.get_deleter()) :
-                                std::unique_ptr<T, Deleter>(rhs.handle_.get(), rhs.handle_.get_deleter());
         return *this;
     }
 
@@ -92,10 +92,38 @@ public:
     }
     
     /**
+     * \brief Clones the internal pointer using the specified cloner function.
+     */
+    T* clone() const {
+        return cloner_ ? cloner_(handle_.get()) : handle_.get();
+    }
+    
+    /**
      * \brief Releases ownership of the internal pointer
      */
     T* release() {
         return handle_.release();
+    }
+    
+    /**
+     * \brief Reset the internal pointer to a new one
+     */
+    void reset(T* ptr) {
+        handle_.reset(ptr);
+    }
+    
+    /**
+     * \brief Get the deleter
+     */
+    const Deleter& get_deleter() const {
+        return handle_.get_deleter();
+    }
+    
+    /**
+     * \brief Get the cloner
+     */
+    const Cloner& get_cloner() const {
+        return cloner_;
     }
     
     /**
--- a/include/cppkafka/clonable_ptr.h
+++ b/include/cppkafka/clonable_ptr.h
@@ -92,13 +92,6 @@ public:
     }
     
     /**
-     * \brief Clones the internal pointer using the specified cloner function.
-     */
-    T* clone() const {
-        return cloner_ ? cloner_(handle_.get()) : handle_.get();
-    }
-    
-    /**
      * \brief Releases ownership of the internal pointer
      */
     T* release() {
@@ -133,6 +126,13 @@ public:
         return static_cast<bool>(handle_);
     }
 private:
+    /**
+     * \brief Clones the internal pointer using the specified cloner function.
+     */
+    T* clone() const {
+        return cloner_ ? cloner_(handle_.get()) : handle_.get();
+    }
+    
     std::unique_ptr<T, Deleter> handle_;
     Cloner cloner_;
 };
--- a/include/cppkafka/clonable_ptr.h
+++ b/include/cppkafka/clonable_ptr.h
@@ -60,7 +60,7 @@ public:
      * \param rhs The pointer to be copied
      */
     ClonablePtr(const ClonablePtr& rhs)
-    : handle_(std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter())),
+    : handle_(std::unique_ptr<T, Deleter>(rhs.try_clone(), rhs.get_deleter())),
       cloner_(rhs.get_cloner()) {
 
     }
@@ -72,7 +72,7 @@ public:
      */
     ClonablePtr& operator=(const ClonablePtr& rhs) {
         if (this != &rhs) {
-            handle_ = std::unique_ptr<T, Deleter>(rhs.clone(), rhs.get_deleter());
+            handle_ = std::unique_ptr<T, Deleter>(rhs.try_clone(), rhs.get_deleter());
             cloner_ = rhs.get_cloner();
         }
         return *this;
@@ -124,10 +124,7 @@ public:
         return static_cast<bool>(handle_);
     }
 private:
-    /**
-     * \brief Clones the internal pointer using the specified cloner function.
-     */
-    T* clone() const {
+    T* try_clone() const {
         return cloner_ ? cloner_(get()) : get();
     }
     
--- a/include/cppkafka/header_list.h
+++ b/include/cppkafka/header_list.h
@@ -192,7 +192,7 @@ bool operator!=(const HeaderList<HeaderType>& lhs, const HeaderList<HeaderType> 
 
 template <typename HeaderType>
 HeaderList<HeaderType> HeaderList<HeaderType>::make_non_owning(rd_kafka_headers_t* handle) {
-    return handle ? HeaderList(handle, NonOwningTag()) : HeaderList();
+    return HeaderList(handle, NonOwningTag());
 }
 
 template <typename HeaderType>
@@ -204,21 +204,19 @@ HeaderList<HeaderType>::HeaderList()
 template <typename HeaderType>
 HeaderList<HeaderType>::HeaderList(size_t reserve)
 : handle_(rd_kafka_headers_new(reserve), &rd_kafka_headers_destroy, &rd_kafka_headers_copy) {
-
+    assert(reserve);
 }
 
 template <typename HeaderType>
 HeaderList<HeaderType>::HeaderList(rd_kafka_headers_t* handle)
-: handle_(handle,
-          handle ? &rd_kafka_headers_destroy : nullptr,
-          handle ? &rd_kafka_headers_copy : nullptr) { //if we own the header list, we clone it on copy
+: handle_(handle, &rd_kafka_headers_destroy, &rd_kafka_headers_copy) { //if we own the header list, we clone it on copy
+    assert(handle);
 }
 
 template <typename HeaderType>
 HeaderList<HeaderType>::HeaderList(rd_kafka_headers_t* handle, NonOwningTag)
-: handle_(handle,
-          handle ? &dummy_deleter : nullptr,
-          nullptr) { //if we don't own the header list, we forward the handle on copy.
+: handle_(handle, &dummy_deleter, nullptr) { //if we don't own the header list, we forward the handle on copy.
+    assert(handle);
 }
 
 // Methods
--- a/include/cppkafka/header_list.h
+++ b/include/cppkafka/header_list.h
@@ -147,12 +147,6 @@ public:
     rd_kafka_headers_t* get_handle() const;
     
     /**
-     * \brief Clone the underlying header list handle.
-     * \return The handle.
-     */
-    rd_kafka_headers_t* clone_handle() const;
-    
-    /**
      * \brief Get the underlying header list handle and release its ownership.
      * \return The handle.
      * \warning After this call, the HeaderList becomes invalid.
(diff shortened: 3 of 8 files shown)


include/cppkafka/clonable_ptr.h
```diff
@@ -91,13 +91,41 @@ class ClonablePtr {
         return handle_.get();
     }
     
+    /**
+     * \brief Clones the internal pointer using the specified cloner function.
+     */
+    T* clone() const {
```
mfontanini: Do you think it makes sense to expose this member function? Shouldn't it be private given you get the same, but safer, behavior if you use the copy constructor?

include/cppkafka/message.h
```diff
@@ -127,6 +127,27 @@ class CPPKAFKA_API Message {
     }
     
 #if (RD_KAFKA_VERSION >= RD_KAFKA_HEADERS_SUPPORT_VERSION)
+    /**
+     * \brief Sets the message's header list.
+     * \note After this call, the Message will take ownership of the header list.
+     */
+    void set_header_list(const HeaderListType& headers) {
```
mfontanini: I don't think this overload is a good idea, at least with the current semantics. If a function takes something by const ref, it should by no means be allowed to take ownership of it. The overload below that takes an rvalue-ref makes sense, as in that case someone specifically moved whatever they were passing. 

Can't this overload just copy the headers?

include/cppkafka/message_builder.h
```diff
@@ -421,17 +470,21 @@ class MessageBuilder : public BasicMessageBuilder<Buffer, MessageBuilder> {
...
     template <typename T>
-    void construct_buffer(Buffer& lhs, const T& rhs) {
-        lhs = Buffer(rhs);
+    void construct_buffer(Buffer& lhs, T&& rhs) {
+        lhs = Buffer(std::forward<T>(rhs));
     }
     
+
     MessageBuilder clone() const {
         return std::move(MessageBuilder(topic()).
```
mfontanini: I didn't spot this when this was first implemented, but the `std::move` here isn't doing anything and shouldn't be there.

include/cppkafka/header_list.h
```diff
@@ -146,6 +146,12 @@ class HeaderList {
      */
     rd_kafka_headers_t* get_handle() const;
     
+    /**
+     * \brief Clone the underlying header list handle.
+     * \return The handle.
+     */
+    rd_kafka_headers_t* clone_handle() const;
```
mfontanini: What is the reason for this member function? Isn't it the equivalent of copying the `HeaderList` via copy constructor?

include/cppkafka/header_list.h
```diff
@@ -254,7 +266,8 @@ HeaderType HeaderList<HeaderType>::at(size_t index) const {
     if (error != RD_KAFKA_RESP_ERR_NO_ERROR) {
         throw Exception(error.to_string());
     }
-    return HeaderType(name, BufferType(value, size));
+    //Use 'Buffer' to implicitly convert to 'BufferType'
+    return HeaderType(name, Buffer(value, size));
```
mfontanini: Is this relying on `Buffer` being implicitly convertible to `BufferType`? Will this work for any `BufferType` that's not `std::string`? I'm not sure what this is achieving here, what's the reason for this change?

include/cppkafka/header_list.h
```diff
@@ -159,10 +165,14 @@ class HeaderList {
      */
     explicit operator bool() const;
     
+    /**
+     * \brief Indicates if this list owns the underlying handle or not.
+     */
+    bool is_owning() const;
```
mfontanini: I don't think this should be exposed. Owning handles can only be created privately so nobody from the outside knows that handles can be non owning. Also see my comment on the place where this is used, as I don't think we even need this (unless there's a use for it outside the library itself).

accelerated:
>A few comments. In the future please either create different PRs or at least split changes into multiple commits, it's not trivial to follow which change is done for which reason here.

Ok but in this case all changes were related.

include/cppkafka/clonable_ptr.h
accelerated: I will make it private. Initially i was using it in the BasicMessageBuilder constructor when you from one type to another. The normal copy construction doesn't work there and you have to work directly by cloning the underlying handle (or calling `rd_kafka_headers_copy`).  Reason I removed it is because there is no way to know if calling `clone()` will be a no-op or a real copy.

include/cppkafka/message_builder.h
accelerated: Negative on this one. The methods return a `Concrete&` which needs explicit move, otherwise you get the copy-constructor invoked, which doesn't even compile actually because `Concrete&` should be `const Concrete&` to call the copy constructor.

include/cppkafka/message.h
accelerated: Well, if you look at the code, I actually assert that it's non-owning. There are instances of `HeaderListType` which are non-owning by default (like the one inside a `Message` class) so in this case passing it by `const` is right. If you only use the rvalue method that means you are moving the handle out, which is not always the case. I can remove this methond (and the `is_owning`) if you like, but consider the above scenario and lmk.

include/cppkafka/header_list.h
accelerated: The reason is explained above in your comment about the `ClonablePtr::clone()`. You can't use the copy constructor if you build a HeaderList<A> from a HeaderList<B>. So this was a way of doing this transformation instead of working with the raw handle. But in the end, since cloning does not guaratee a copy, I will remove it.

include/cppkafka/header_list.h
accelerated: See my comment above. I can remove it if you like.

include/cppkafka/header_list.h
accelerated: Yes, exactly. Also there's the conversion into `vector<unsigned char>` which is what I needed in my case. If I don't use this conversion, I pretty much have to come up with some private methods templated on the `BufferType` in which case I would do a string and vector one, so might as well use the conversion operator. 

Also `string` class has a constructor which is (ptr, size), but `vector` only has (iter_start, iter_end). Alternatively I tried adding a `Buffer::Buffer(iter_start, iter_end)` constructor so that I can skip the conversion operator, but that gives all sorts of compiler errors due to a clash with its (ptr, size) constructor. `string` class uses some SFINEA tricks for that I believe so it was too much work in this case to make it compile. Again you decide what you prefer.

include/cppkafka/message_builder.h
mfontanini: Ah, I see. I'd usually prefer constructing an object, calling everything and then just returning it (no need for move there) but that's okay.

include/cppkafka/message.h
mfontanini: So currently the only place where there's non owning headers is in a `Message`, right? This means that the only way to use this overload is if you grab a produced message, let's call it A, and message B which for example just failed to be produced and you want to re-produce it using the headers from message A. Am I missing something? 

You can only use this overload if you pull non owning headers from somewhere and those currently are only inside a Message. Either that or someone explicitly created a `rd_kafka_headers_t` and built a non owning cppkafka wrapper around it and now want to use it.

Given these use cases (please correct me if I'm missing something) I think:
* The use case of pulling headers from a message and putting them into another one is not common at all so people can just use the other overload.
* We shouldn't be encouraging people to use non owning handles because they make everything more complicated. It's better if they stick to what cppkafka provides or if there's extra functionality missing, the library should provide it.

I'd even go ahead and merge both overloads into one that simply takes a `HeaderListType` by value. If you want to move your own in there, do it, otherwise it gets copied.

include/cppkafka/header_list.h
mfontanini: I take back the "can only be created privately", that's not true. Anyway, let's follow this on the other comment.

include/cppkafka/message.h
accelerated: Ok I changed the overload so that when passing by `const HeaderList&` we actually make a copy via `rd_kafka_header_copy` and don't enfoce ownership. Merging the two and passing it by value does not guarantee that the _copy_ will actually contain a copied list. If the HeaderList is non-owning, the copy is nothing more than copying the handle around, which we then pass ownership to rdkafka, so whoever calls this setter method will get shot in the foot w/o knowing. 
Yes, in reference to your comment, the only non-owning instance is inside Message, however we can also make `MessageBuilder` non-owning as per my comments regarding the excessive copies. Also you never know what the users might do. They might call `get_handle()` on a message builder or message, make a new non-owning HeaderList and then pass that around. So we cannot assume that only a Message contains non-owning headers.

accelerated:
@mfontanini, please check the last few commits. I didn't collapse them so that you can see each one individually. 
Also you haven't really addressed the multiple copy issue...such as making `MessageBuilder` non-owning? or perhaps allowing `BufferedProducer` and Producer to produce with a `ConcreteMessageBuilder` directly?

include/cppkafka/header_list.h
```diff
@@ -260,8 +260,7 @@ HeaderType HeaderList<HeaderType>::at(size_t index) const {
     if (error != RD_KAFKA_RESP_ERR_NO_ERROR) {
         throw Exception(error.to_string());
     }
-    //Use 'Buffer' to implicitly convert to 'BufferType'
-    return HeaderType(name, Buffer(value, size));
+    return HeaderType(name, BufferType(value, value + size));
```
mfontanini: Nice, this looks better

include/cppkafka/clonable_ptr.h
```diff
@@ -98,13 +96,41 @@ class ClonablePtr {
...
      */
     explicit operator bool() const {
         return static_cast<bool>(handle_);
     }
 private:
+    /**
+     * \brief Clones the internal pointer using the specified cloner function.
+     */
+    T* clone() const {
+        return cloner_ ? cloner_(get()) : get();
```
mfontanini: If there's no cloner, this should probably throw, right? Otherwise you think you got a new pointer but in fact it's the same as the old one and you'll end up deleting it twice.

include/cppkafka/header_list.h
```diff
@@ -210,14 +209,16 @@ HeaderList<HeaderType>::HeaderList(size_t reserve)
 
 template <typename HeaderType>
 HeaderList<HeaderType>::HeaderList(rd_kafka_headers_t* handle)
-: handle_(handle, &rd_kafka_headers_destroy, &rd_kafka_headers_copy) { //if we own the header list, we clone it on copy
-
+: handle_(handle,
+          handle ? &rd_kafka_headers_destroy : nullptr,
+          handle ? &rd_kafka_headers_copy : nullptr) { //if we own the header list, we clone it on copy
```
mfontanini: I'm not sure about this change. Why do we want to allow building a `HeaderList` on top of a null handle? This should probably assert that the handle isn't null. I presume you're using this in `Message::set_header_list`, right? I'd rather than one make sure it's not calling this with a null pointer rather than this one "adapting" to null pointers when in reality this shouldn't work.

include/cppkafka/header_list.h
```diff
@@ -210,14 +209,16 @@ HeaderList<HeaderType>::HeaderList(size_t reserve)
...
 }
 
 template <typename HeaderType>
 HeaderList<HeaderType>::HeaderList(rd_kafka_headers_t* handle, NonOwningTag)
-: handle_(HandlePtr(handle, &dummy_deleter, &dummy_cloner)) { //if we don't own the header list, we forward the handle on copy.
-
+: handle_(handle,
+          handle ? &dummy_deleter : nullptr,
```
mfontanini: Same as the comment above.

include/cppkafka/clonable_ptr.h
accelerated: Well this function is now private so its only purpose is to greatly simplify the constructor and copy constructor code  (see how it was written originally). If it was public I would agree with you.

accelerated: We can call it something else than `clone()` if you prefer, maybe `copy` or `mirror`?

include/cppkafka/message_builder.h
accelerated: I will change it as you like.

include/cppkafka/header_list.h
accelerated: ok

accelerated:
Please check my latest commit and lmk is it's ok.

include/cppkafka/clonable_ptr.h
mfontanini: I don't think it's a matter of naming. The only case where you won't have a cloner function is if the `ClonablePtr` was moved, right? In that case, you won't have a pointer either. So this boils down to how you want to handle copy constructing/assigning from an empty `ClonablePtr`. I don't imagine all rdkafka destroy functions will work correctly if you pass in a null pointer, right? In which case, this shouldn't actually be returning just `get()` as that would end up blowing up later, given `get()` is a null pointer.

The result of copy constructing from an empty `ClonablePtr` should be that you have an empty pointer with an empty deleter (which is actually the same state `rhs` is in). A combination of pointer + non null deleter will probably go :boom:.

accelerated: Yeah well the times when you have a null cloner function is when 1) the `ClonablePtr` was moved and 2) the `ClonablePtr` was null to begin with - aka built via default `HeaderList` constructor. Whenever I use directly the rdkafka destroy or copy methods (which require a valid handle!) I always check if the `HeaderList` object is set (i.e. via bool operator). If not I never call the rdkafka functions. The only time when you can pass a null header handle is in the HEADERS_V macro inside `Producer::produce` because from what I can see i the rdkafka code, the headers are always checked before they are dereferenced, so in this case you can pass a null ptr. 
The combo of non-null pointer with null deleter can't exist...not sure where you see that?

accelerated: This is a sample code from rdkafka:
```
void rd_kafka_headers_destroy (rd_kafka_headers_t *hdrs) {
        rd_list_destroy(&hdrs->rkhdrs_list); <---- BOOM
        rd_free(hdrs);
}
```
I personally would have put a guard around this logic, just like calling `delete nullptr` doesn't blow up.

include/cppkafka/header_list.h
mfontanini: The thing is, this constructor will be called both by cppkafka and external users of the library. You want to be friendly to external users and not let their broken code compile (e.g. allowing doing `HeaderList(nullptr)` shouldn't work)

If there's 3 places inside cppkafka where there's the possibility of calling that with an null pointer, I'd rather add the checks in those 3 places rather than being slightly more flexible on the `HeaderList` API but allowing the user to create a header that contains nothing bypassing the default constructor.

include/cppkafka/clonable_ptr.h
mfontanini: Well, we can't change rdkafka's code so we have to deal with it. As it is now, if you construct a `ClonablePtr` using an empty cloner and then copy construct from it, it blows up. This shouldn't blow up and instead should just throw.

include/cppkafka/header_list.h
accelerated: Yes, this is what I've done in the last commit.

include/cppkafka/clonable_ptr.h
accelerated: I'm not sure if we're out of sync here...the only place where the cloner is called is inside the private method `ClonablePtr::try_clone()` (I renamed it since it does not always clone) and the validity of it is checked before being invoked. In the case of the copy constructor with an empty cloner (this is the standard non-owning case) it works perfectly fine, the handle is just copied as-is along with the DummyDeleter function. Don't think it will blow up. Unless I'm missing something...

include/cppkafka/message_builder.h
```diff
@@ -234,23 +243,50 @@ template <typename T, typename C>
...
 }
 
+template <typename T, typename C>
+template <typename U, typename V>
+BasicMessageBuilder<T, C>::BasicMessageBuilder(BasicMessageBuilder<U, V>&& rhs)
+: topic_(rhs.topic()),
+  partition_(rhs.partition()),
+#if (RD_KAFKA_VERSION >= RD_KAFKA_HEADERS_SUPPORT_VERSION)
+  header_list_(rhs.header_list() ?
+               HeaderListType(rhs.header_list().release_handle()) : HeaderListType()), //assume header ownership
```
mfontanini: Isn't it cleaner to just do `header_list_(move(rhs.header_list()))`? `HeaderList`s is movable, right?

include/cppkafka/message_builder.h
```diff
@@ -234,23 +243,50 @@ template <typename T, typename C>
 BasicMessageBuilder<T, C>::BasicMessageBuilder(const Message& message)
 : topic_(message.get_topic()),
   key_(Buffer(message.get_key().get_data(), message.get_key().get_size())),
+#if (RD_KAFKA_VERSION >= RD_KAFKA_HEADERS_SUPPORT_VERSION)
+  header_list_(message.get_header_list() ?
+               HeaderListType(rd_kafka_headers_copy(message.get_header_list().get_handle())) : HeaderListType()), //copy headers
```
mfontanini: Similar to the comment about moving, `HeaderList` is copyable, right? Does this achieve something different than what `header_list_(message.get_header_list())` does? Maybe I'm missing something.

include/cppkafka/message_builder.h
```diff
@@ -234,23 +243,50 @@ template <typename T, typename C>
...
 
 template <typename T, typename C>
 template <typename U, typename V>
 BasicMessageBuilder<T, C>::BasicMessageBuilder(const BasicMessageBuilder<U, V>& rhs)
-: topic_(rhs.topic()), partition_(rhs.partition()), timestamp_(rhs.timestamp()),
+: topic_(rhs.topic()),
+  partition_(rhs.partition()),
+#if (RD_KAFKA_VERSION >= RD_KAFKA_HEADERS_SUPPORT_VERSION)
+  header_list_(rhs.header_list() ?
+               HeaderListType(rd_kafka_headers_copy(rhs.header_list().get_handle())) : HeaderListType()), //copy headers
```
mfontanini: Same comment as above regarding copying.

include/cppkafka/clonable_ptr.h
mfontanini: Well, it will blow up if the deleter is not dummy. This is used internally though so it should be fine.

include/cppkafka/buffer.h
```diff
@@ -81,6 +81,17 @@ class CPPKAFKA_API Buffer {
...
     }
+    
+    /**
+     * Constructs a buffer from two iterators in the range [first,last)
+     *
+     * \param first An iterator to the start of data
+     * \param last An iterator to the end of data (not included)
+     */
+    template <typename Iter>
+    Buffer(const Iter first, Iter last)
```
mfontanini: Both of these can be `const`. I'd also add the same `static_assert` that the other constructors have, using `iterator_traits` to check what the contained type is.

include/cppkafka/header_list.h
```diff
@@ -281,18 +279,13 @@ bool HeaderList<HeaderType>::empty() const {
 template <typename HeaderType>
 typename HeaderList<HeaderType>::Iterator
 HeaderList<HeaderType>::begin() const {
-    assert(handle_);
-    if (empty()) {
-        return end();
-    }
-    return Iterator(make_non_owning(handle_.get()), 0);
+    return empty() ? end() : Iterator(make_non_owning(handle_.get()), 0);
```
mfontanini: What's the reason why this can't just return `Iterator(make_non_owning(handle_.get()), 0)`? Everything should be internally checking if the handle is valid or not at this point. Same for `end` below.

include/cppkafka/message_builder.h
accelerated: In this function I can't just call the move or the copy constructor because you're copying/moving from `HeaderList<Header<T>>` to `HeaderList<Header<U>>` and there's no conversion constructor so you have to pass via the raw handle. Otherwise I can implement a copy constructor from `HeaderList<A>` to `HeaderList<B>` in which case you can then use the move or copy directly. Let me know if you prefer that...

include/cppkafka/buffer.h
mfontanini: Nevermind, I didn't realize this was forwarding. Ignore my second comment.

include/cppkafka/message_builder.h
accelerated: No, because `Message` is a non-owning object so if you move the headers, MessageBuilder will now have a non-owning handle as well. When `Message` goes out of scope, rdkafka deletes the header handle so the `MessageBuilder` is now left with a bad handle. So the only way is once again to do a raw copy of the handle.

mfontanini: That's correct for the constructor overload that takes a message, but this one takes a message builder. What you're doing here is equivalent to moving, it's just that you're doing manually.

include/cppkafka/header_list.h
accelerated: Because you can't call `make_non_owning(null)` so if your handle is null, it will assert. This is based on your preference to not adapt the HeaderList to a null/non-null handle.

accelerated:
I think much of the complexity here stems from the fact that when you have a non-owning header list which you _copy/move_ what is the expected result ? Do you expect an exact copy (i.e. another non-owning header list - which is the current behavior) or do you expect a real copy of the underlying kafka handle? In that case the `ClonerPtr` may not be the best class to use for this.

include/cppkafka/header_list.h
mfontanini: Touche. Alright, let's keep it as it is then.

mfontanini:
Yeah, I agree. This mix of non owning handles that when copied have to become owned is not really what `ClonablePtr` was meant to be used for. It may be worth adding a thing RAII'd wrapper on top of a `rd_kafka_headers_t*` that provides the right semantics, regardless of whether it's owned or not. Do you think that would remove the need for handling these corner cases?

accelerated:
@mfontanini Ok I made another round of code cleanup. I this will be as close as we can get to what you were targeting.

accelerated:
> Nice, this looks much better

Ok, lmk if there's anything else you want me to address for this PR.

mfontanini:
Alright this looks good. Looks like tests are failing when starting zookeeper but I ran them locally and they all succeed. Thanks for the PR!

accelerated:
Cheers!!

PR #140, opened 2018-11-20
Add support for Buffer construction via raw arrays

tgaldes:

PR #147, opened 2018-12-10
Support for raw array Buffer constructor

accelerated:
Finished https://github.com/mfontanini/cppkafka/pull/140


mfontanini:
Thanks!

PR #148, opened 2018-12-13
Added flush/produce termination callbacks

accelerated:
Added `BufferedProducer::FlushTerminationCallback`. 

Reason:
1) The current `FlushFailureCallback` can be called multiple times subject to the number of retries set. However, the application has no way of knowing how many retries are left (the retry count is not passed into the `FlushFailureCallback`) so unless it keeps track via a map or some other logic of each message retry count it cannot detect when a message will be dropped. This makes message accounting tedious and complex. 
2) Second reason is that the application needs to know when a message delivery fails. In this case, the `ProduceFailureCallback` cannot be called because we cannot build a `Message` object w/o a valid rdkafka handle. The application may also need to do some cleanup of the `opaque` data it has passed. Otherwise this data can be leaked. Also the application can keep stats of how many messages fail or perhaps re-publish them into another kafka queue, log them, etc. And for that the `FlushFailureCallback` is insufficient. 
3) Same logic applies to the `ProduceFailureCallback` vs `ProduceTerminationCallback` pair.

### DIFF
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -104,15 +104,26 @@ public:
      * Callback to indicate a message failed to be produced by the broker.
      *
      * The returned bool indicates whether the BufferedProducer should try to produce
-     * the message again after each failure.
+     * the message again after each failure, subject to the maximum number of retries set. If this callback
+     * is not set or returns false or if the number of retries reaches zero, the ProduceTerminationCallback
+     * will be called.
      */
     using ProduceFailureCallback = std::function<bool(const Message&)>;
+    
+    /**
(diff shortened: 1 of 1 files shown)


demin80:
This would be quite useful for our apps! If a message is dropped by cppkafka, the apps need to be notified.

mfontanini:
Sorry I didn't reply, I'll check this (and the other(s) PR(s)) over the weekend.

accelerated:
> Sorry I didn't reply, I'll check this (and the other(s) PR(s)) over the weekend.

No rush, I'm off for a few weeks so I won't be able to fix your comments anyway. But I'll check when I get back in Jan. Some other ppl at work are also waiting for these PRs so they will most likely code around them in the meantime.

PR #149, opened 2018-12-13
Added queue full notification

accelerated:
Added queue full notification with various triggering options. This is needed so that the application can be notified when the rdkafka queue is backed up and can be allowed to reduce upstream pressure on the producers.

### DIFF
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -92,9 +92,9 @@ public:
         Async    ///< Empty the buffer and don't wait for acks.
     };
     enum class QueueFullNotification {
-        None,          ///< Don't notify
-        EdgeTriggered, ///< Notify once. Application must call queue_full_trigger_reset() to enable again.
-        EachOccurence  ///< Notify on each occurence.
+        None,           ///< Don't notify (default).
+        OncePerMessage, ///< Notify once per message.
+        EachOccurence   ///< Notify on each occurence.
     };
     /**
      * Concrete builder
@@ -144,6 +144,14 @@ public:
      * then FlushFailureCallback should not be set.
      */
     using FlushTerminationCallback = std::function<void(const MessageBuilder&, Error error)>;
+    
+    /**
+     * Callback to indicate a queue full error was received when producing.
+     *
+     * The MessageBuilder instance represents the message which triggered the error. This callback will be called
+     * according to the set_queue_full_notification() setting.
+     */
+    using QueueFullCallback = std::function<void(const MessageBuilder&)>;
 
     /**
      * \brief Constructs a buffered producer using the provided configuration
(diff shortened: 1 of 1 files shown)


include/cppkafka/utils/buffered_producer.h
```diff
@@ -772,6 +819,13 @@ void BufferedProducer<BufferType, Allocator>::produce_message(BuilderType&& buil
             if (ex.get_error() == RD_KAFKA_RESP_ERR__QUEUE_FULL) {
                 // If the output queue is full, then just poll
                 producer_.poll();
+                // Notify application so it can slow-down production
+                if (queue_full_notify) {
+                    queue_full_notify = queue_full_trigger_ = false; //clear trigger and local state
```
mfontanini: This is always resetting `queue_full_notify`, so if you actually use `QueueFullNotification::EachOccurence` you'll still get only once per `produce_message` call, right? Not sure if the issue is here or if you wanted "each occurrence" to still mean "every time but only once per message"/

accelerated:
Yes that's the idea...`EachOccurence` is once per message, otherwise you can get spammed by too many callbacks.

mfontanini:
Right but what if you keep hitting the limit forever for a single message? There's currently no way of actually knowing you get stuck. "Each occurrence" makes me think it's _actually_ every occurrence of a full queue rather than "only one occurrence per message".

accelerated:
Yes, there's no way of knowing if you're stuck or even which message got stuck. It's better than being silently loopinh though. Another alternative would be to also add one enum for "every single time" which would essentially be for each error multiple times per message, if it so happens. And instead of using the error callback we can have a separate callback just for this, with an argument taking a MessageBuilder reference, so the application can also see what got stuck. Otherwise, if you have other suggestions i'd be glad to implement them.

So then:
`EdgeTriggered` -> fire once with manual reset
`OncePerMessage` -> fire once per message.
`EachOccurence` -> every time it happens.

accelerated:
Hi @mfontanini, i am ready to resume this. Can you please lmk what changes you want? Thanks!

accelerated:
Ok that makes sense. Will make changes as you requested. Stay tuned.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -358,6 +373,18 @@ class CPPKAFKA_API BufferedProducer {
      * Simple helper to construct a builder object
      */
     Builder make_builder(std::string topic);
+    
+    /**
+     * Set the type of notification when RD_KAFKA_RESP_ERR__QUEUE_FULL is received.
+     *
+     * This will call the error callback for this producer. By default this is set to QueueFullNotification::None.
+     */
+    void set_queue_full_notification(QueueFullNotification notification);
```
mfontanini: Maybe the getter for the notifiaction enum and the callback should be done in just one? You'll always call either both or none of them so I think that would make sense.

accelerated: You mean the setter of the notification type? I think there might be cases when the app wants to set the notifocation to some enum and then it might want to turn it off (None) for a period and re-enable it later. In that case it would have to set the callback each time...prob not desirable.

mfontanini: Alright, fair enough. Merging this one, thanks!

PR #153, opened 2019-01-07
Added a high-priority queue to BufferedProducer to avoid message re-ordering

demin80:
Currently, when a **BufferedProducer** is used, and produced messages are failed to be sent (async_produce) or failed to be delivered (on_delivery_report), then **do_add_message** will enqueue the messages to the messsages_ queue with the **MessagePriority::High** priority. However, the existing code enqueues them to the front of the queue via emplace_front. This can cause reversing the original order of messages: e.g., failing to send messages M1, M2, M3 will result in having M3, M2, M1 in messages_. So if the next attempt to flush the queue succeeds, the messages will be sent in the reverse order w.r.t. the original. This pull request adds a high-priority message queue to BufferedProducer: high-priority messages are always enqueued to the back of the queue, and this queue is always flushed before the low-priority message queue.

### DIFF
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -530,6 +530,7 @@ private:
     // Members
     Producer producer_;
     QueueType messages_;
+    QueueType hi_pri_messages_;
     mutable std::mutex mutex_;
     ProduceSuccessCallback produce_success_callback_;
     ProduceFailureCallback produce_failure_callback_;
@@ -565,7 +566,8 @@ template <typename BufferType, typename Allocator>
 BufferedProducer<BufferType, Allocator>::BufferedProducer(Configuration config,
                                                           const Allocator& alloc)
 : producer_(prepare_configuration(std::move(config))),
-  messages_(alloc) {
+  messages_(alloc),
+  hi_pri_messages_(alloc) {
     producer_.set_payload_policy(get_default_payload_policy<BufferType>());
 #ifdef KAFKA_TEST_INSTANCE
     test_params_ = nullptr;
@@ -625,10 +627,16 @@ template <typename BufferType, typename Allocator>
 void BufferedProducer<BufferType, Allocator>::async_flush() {
     CounterGuard<size_t> counter_guard(flushes_in_progress_);
     QueueType flush_queue; // flush from temporary queue
+    QueueType hi_pri_flush_queue; // flush from hi-priority temporary queue
     {
         std::lock_guard<std::mutex> lock(mutex_);
+        std::swap(hi_pri_messages_, hi_pri_flush_queue);
         std::swap(messages_, flush_queue);
     }
+    while (!hi_pri_flush_queue.empty()) {
+        async_produce(std::move(hi_pri_flush_queue.front()), false);
+        hi_pri_flush_queue.pop_front();
+    }    
     while (!flush_queue.empty()) {
         async_produce(std::move(flush_queue.front()), false);
         flush_queue.pop_front();
@@ -640,10 +648,16 @@ void BufferedProducer<BufferType, Allocator>::flush(bool preserve_order) {
     if (preserve_order) {
         CounterGuard<size_t> counter_guard(flushes_in_progress_);
         QueueType flush_queue; // flush from temporary queue
+        QueueType hi_pri_flush_queue; // flush from hi-priority temporary queue
         {
             std::lock_guard<std::mutex> lock(mutex_);
+            std::swap(hi_pri_messages_, hi_pri_flush_queue);
             std::swap(messages_, flush_queue);
         }
+        while (!hi_pri_flush_queue.empty()) {
+            sync_produce(hi_pri_flush_queue.front());
+            hi_pri_flush_queue.pop_front();
+        }        
         while (!flush_queue.empty()) {
             sync_produce(flush_queue.front());
             flush_queue.pop_front();
@@ -661,25 +675,45 @@ bool BufferedProducer<BufferType, Allocator>::flush(std::chrono::milliseconds ti
     if (preserve_order) {
         CounterGuard<size_t> counter_guard(flushes_in_progress_);
         QueueType flush_queue; // flush from temporary queue
+        QueueType hi_pri_flush_queue; // flush from hi-priority temporary queue
         {
             std::lock_guard<std::mutex> lock(mutex_);
+            std::swap(hi_pri_messages_, hi_pri_flush_queue);
             std::swap(messages_, flush_queue);
         }
         auto remaining = timeout;
-        auto start_time = std::chrono::high_resolution_clock::now();
+        auto start_time = std::chrono::high_resolution_clock::now();        
         do {
-            sync_produce(flush_queue.front());
-            flush_queue.pop_front();
+            if (!hi_pri_flush_queue.empty()) {
+                sync_produce(hi_pri_flush_queue.front());
+                hi_pri_flush_queue.pop_front();
+            }
+            else if (!flush_queue.empty()) {
+                sync_produce(flush_queue.front());
+                flush_queue.pop_front();
+            }
+            else {
+                break;
+            }
             // calculate remaining time
             remaining = timeout - std::chrono::duration_cast<std::chrono::milliseconds>
                 (std::chrono::high_resolution_clock::now() - start_time);
-        } while (!flush_queue.empty() && (remaining.count() > 0));
+        } while (remaining.count() > 0);
         
         // Re-enqueue remaining messages in original order
-        if (!flush_queue.empty()) {
+        if (!hi_pri_flush_queue.empty() || !flush_queue.empty()) {
             std::lock_guard<std::mutex> lock(mutex_);
-            messages_.insert(messages_.begin(), std::make_move_iterator(flush_queue.begin()), std::make_move_iterator(flush_queue.end()));
-        }
+            if (!!hi_pri_flush_queue.empty()) {
+                hi_pri_messages_.insert(hi_pri_messages_.begin(),
+                                        std::make_move_iterator(hi_pri_flush_queue.begin()),
+                                        std::make_move_iterator(hi_pri_flush_queue.end()));
+            }
+            if (!flush_queue.empty()) {
+                messages_.insert(messages_.begin(),
+                                 std::make_move_iterator(flush_queue.begin()),
+                                 std::make_move_iterator(flush_queue.end()));
+            }
+        }        
     }
     else {
         async_flush();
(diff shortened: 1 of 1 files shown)


accelerated:
This is definitely an issue with the current code. The proposed changes look good to me...

include/cppkafka/utils/buffered_producer.h
```diff
@@ -661,25 +675,45 @@ bool BufferedProducer<BufferType, Allocator>::flush(std::chrono::milliseconds ti
...
-        } while (!flush_queue.empty() && (remaining.count() > 0));
+        } while (remaining.count() > 0);
         
         // Re-enqueue remaining messages in original order
-        if (!flush_queue.empty()) {
+        if (!hi_pri_flush_queue.empty() || !flush_queue.empty()) {
             std::lock_guard<std::mutex> lock(mutex_);
-            messages_.insert(messages_.begin(), std::make_move_iterator(flush_queue.begin()), std::make_move_iterator(flush_queue.end()));
-        }
+            if (!!hi_pri_flush_queue.empty()) {
```
mfontanini: There's a double `!` here, this is actually being executed if the queue is empty.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -774,13 +810,14 @@ void BufferedProducer<BufferType, Allocator>::do_add_message(BuilderType&& build
...
-            messages_.emplace_front(std::forward<BuilderType>(builder));
+            hi_pri_messages_.emplace_back(std::forward<BuilderType>(builder));
         }
         else {
             messages_.emplace_back(std::forward<BuilderType>(builder));
         }
     }
-    if (do_flush && (max_buffer_size_ >= 0) && (max_buffer_size_ <= (ssize_t)messages_.size())) {
+
+    if (priority == MessagePriority::Low && do_flush && (max_buffer_size_ >= 0) && (max_buffer_size_ <= get_buffer_size())) {
```
mfontanini: Can you add a comment in the code on why this is being done? If I read this without knowing what priorities mean in this context, this is confusing/counter intuitive: "if you're adding a message with high priority, then don't flush. Only do that if it's low priority". A comment would clarify that.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -530,6 +530,7 @@ class CPPKAFKA_API BufferedProducer {
     // Members
     Producer producer_;
     QueueType messages_;
+    QueueType hi_pri_messages_;
```
mfontanini: I'm not a fan of abbreviations. I'd rather this be "high_priority_messages_" rather than "hi_pri" which is not very self describing.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -625,10 +627,16 @@ template <typename BufferType, typename Allocator>
 void BufferedProducer<BufferType, Allocator>::async_flush() {
     CounterGuard<size_t> counter_guard(flushes_in_progress_);
     QueueType flush_queue; // flush from temporary queue
+    QueueType hi_pri_flush_queue; // flush from hi-priority temporary queue
```
mfontanini: Same with "hi_pri" here

include/cppkafka/utils/buffered_producer.h
```diff
@@ -625,10 +627,16 @@ template <typename BufferType, typename Allocator>
...
     {
         std::lock_guard<std::mutex> lock(mutex_);
+        std::swap(hi_pri_messages_, hi_pri_flush_queue);
         std::swap(messages_, flush_queue);
     }
+    while (!hi_pri_flush_queue.empty()) {
```
mfontanini: I'd probably create a private function `async_produce(QueueType queue, bool throw_on_error)` and call it twice to avoid code duplication. Otherwise at least create a local lambda and call it with each queue, that also works.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -640,10 +648,16 @@ void BufferedProducer<BufferType, Allocator>::flush(bool preserve_order) {
     if (preserve_order) {
         CounterGuard<size_t> counter_guard(flushes_in_progress_);
         QueueType flush_queue; // flush from temporary queue
+        QueueType hi_pri_flush_queue; // flush from hi-priority temporary queue
         {
             std::lock_guard<std::mutex> lock(mutex_);
+            std::swap(hi_pri_messages_, hi_pri_flush_queue);
             std::swap(messages_, flush_queue);
         }
+        while (!hi_pri_flush_queue.empty()) {
```
mfontanini: Same comment avoid code duplication here.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -661,25 +675,45 @@ bool BufferedProducer<BufferType, Allocator>::flush(std::chrono::milliseconds ti
...
+        } while (remaining.count() > 0);
         
         // Re-enqueue remaining messages in original order
-        if (!flush_queue.empty()) {
+        if (!hi_pri_flush_queue.empty() || !flush_queue.empty()) {
             std::lock_guard<std::mutex> lock(mutex_);
-            messages_.insert(messages_.begin(), std::make_move_iterator(flush_queue.begin()), std::make_move_iterator(flush_queue.end()));
-        }
+            if (!!hi_pri_flush_queue.empty()) {
+                hi_pri_messages_.insert(hi_pri_messages_.begin(),
```
mfontanini: Same comment avoid code duplication as well.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -735,11 +769,13 @@ void BufferedProducer<BufferType, Allocator>::clear() {
     std::lock_guard<std::mutex> lock(mutex_);
     QueueType tmp;
     std::swap(tmp, messages_);
+    QueueType hi_pri_tmp;
```
mfontanini: "hi_pri" again here

demin80:
Thank you guys for your comments. I've updated the implementation and hopefully addressed all the suggestions received so far:
1) Removed the code duplication via lambdas;
2) Renamed the abbreviated hi_pri_messages_ to retry_messages_;
3) Replaced MessagePriority with a boolean is_retry, as MessagePriority::High means "retry sending"
4) Added the comment explaining why we're not flushing the queue when a retry message is added. Essentially, queue flushing involves locking the mutex, and it's not a good idea to lock the producer in an rdkafka callback (which is where replay messages may be added from).

include/cppkafka/utils/buffered_producer.h
```diff
@@ -661,25 +672,44 @@ bool BufferedProducer<BufferType, Allocator>::flush(std::chrono::milliseconds ti
...
+                return true;
+            }
+            return false;
+        };
         auto remaining = timeout;
         auto start_time = std::chrono::high_resolution_clock::now();
         do {
-            sync_produce(flush_queue.front());
-            flush_queue.pop_front();
+            if (!queue_flusher(retry_flush_queue) && !queue_flusher(flush_queue)) {
```
accelerated: I think this statement should be `!(queue_flusher(retry_flush_queue) || queue_flusher(flush_queue))` because the way it's written now both flushers will be evaluated which means you will flush alternatively from each queue. The retry queue should be fully flushed _before_ the main queue.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -661,25 +672,44 @@ bool BufferedProducer<BufferType, Allocator>::flush(std::chrono::milliseconds ti
...
         
         // Re-enqueue remaining messages in original order
-        if (!flush_queue.empty()) {
-            std::lock_guard<std::mutex> lock(mutex_);
-            messages_.insert(messages_.begin(), std::make_move_iterator(flush_queue.begin()), std::make_move_iterator(flush_queue.end()));
-        }
+        auto re_enqueuer = [this](QueueType& src_queue, QueueType& dst_queue)->void
+        {
+            if (!src_queue.empty()) {
+                std::lock_guard<std::mutex> lock(mutex_);
```
accelerated: Since the logic for flushing/re-enqueuing has now been made generic into lambdas, I wonder if it's not better to have 2 mutexes (one for each queue) and just pass it in the lambda as well as a parameter. There is no point in blocking operations on one queue when the other one is being accessed. Especially since the `retry_queue_` is populated primarily from the rdkafka callback, you don't want to block it unnecessarily when the main queue is being used.

include/cppkafka/utils/buffered_producer.h
```diff
@@ -519,17 +518,18 @@ class CPPKAFKA_API BufferedProducer {
         return nullptr;
     }
     template <typename BuilderType>
-    void do_add_message(BuilderType&& builder, MessagePriority priority, bool do_flush);
+    void do_add_message(BuilderType&& builder, bool is_retry, bool do_flush);
```
accelerated: [suggestion for readability] I think that using too many bools in the function signature makes it hard to understand when you read code. What is `do_add_message(builder, true, false)`? or `do_add_message(builder, false, false)`? difficult to tell unless you know the parameters beforehand. I would be ideal to have enums and use something like:
`do_add_message(builder, QueueType::RetryQueue, FlushAction::flush)`

This would also apply to `void async_produce(BuilderType&& message, bool throw_on_error)` where the `throw_on_error` would be made into an enum like `ThrowAction::OnError` or something like that. I would leave this to @demin80 to decide.

include/cppkafka/utils/buffered_producer.h
demin80: The condition statements "(!queue_flusher(retry_flush_queue) && !queue_flusher(flush_queue)" and "!(queue_flusher(retry_flush_queue) || queue_flusher(flush_queue))" are equivalent. 

In the former condition, if queue_flusher(retry_flush_queue) returns true, then the negation of it will evaluate to false, hence because of the short-circuit implementation, the second part of the condition queue_flusher(flush_queue) will not be evaluated because it's clear that the entire statement will evaluate to false regardless. So queue_flusher(flush_queue) will be evaluated if and only if queue_flusher(retry_flush_queue) returns false.

However, if you think that changing the way you suggested will improve the readability, I'm ready to do that.

accelerated: Yes you're right, that makes sense. I somehow only thought of the false case.

demin80:
As pointed by **mfontanini**:
1) The spacing was fixed;
and as suggested by **accelerated**:
2) retry_mutex_ was added to protect retry_messages_;
3) Some bools were replaced with enums.

accelerated:
LGTM

demin80:
Any more comments here?

mfontanini:
I haven't had time to look at this again in detail. I'll have a look tonight and merge it if it looks good.

demin80:
Thanks Matias!

mfontanini:
Thanks for the fix!

PR #156, opened 2019-01-12
Use CMAKE_STATIC_LIBRARY_* macros for FindRdKafka

psigen:
In the current implementation, library suffixes are hard coded from a hand-maintained list.  Instead of writing this list, we can use the CMake macros for platform specific library prefix/suffixes.

E.g. https://cmake.org/cmake/help/v3.0/variable/CMAKE_STATIC_LIBRARY_SUFFIX.html

This also resolves library resolution on Mac OSX, which does not currently work on the native `.dylib`  suffix for shared libraries.


psigen:
I don't think it should, because it has the same search order as before
(specific filename before general filename).

It's been working fine when doing static linking for me, but of course I
would welcome other testers.

mfontanini:
Nice, I didn't know those existed. I don't think this should break anything, it should be equivalent while also working in cases like OSX. Thanks!

PR #157, opened 2019-01-16
added type conversion to avoid signed-vs-unsigned-comparison warning

demin80:
added type conversion to avoid signed-vs-unsigned-comparison warning


demin80:
https://github.com/mfontanini/cppkafka/pull/153 introduced a compilation warning due to the comparison of signed and unsigned values. This PR fixes it by converting the type to ssize_t as it was before.

accelerated:
Also the block must return true [here](https://github.com/mfontanini/cppkafka/pull/157/files#diff-3dc154855924dbfed7813e7e64ad970cL711)

mfontanini:
Thanks! And good catch @accelerated !

PR #164, opened 2019-02-04
Added consumer legacy offset store API

accelerated:
Added support for legacy rdkafka `offset_store()` and `offsets_store()` APIs. These APIs are still very much needed if the client application wants higher performance throughput when committing. As indicated by the rdkafka author himself, the `offset_store()` APIs should be used to avoid congestion if producing/consuming at a high rate. See description of issue [here](https://github.com/edenhill/librdkafka/wiki/FAQ#why-committing-each-message-is-slow) and the suggested config [here](https://github.com/edenhill/librdkafka/wiki/Consumer-offset-management#configuration-properties).
We have also experienced significant slowdown when using the individual `commit()` (even in async mode) and switching to the `offset_store()` API remedied that.


matias-te:
I'm fine with adding this but I'm curious why you need this type of API. Are you committing one message at a time and that (understandably) slows you down?

mfontanini:
I'm fine with adding this but I'm curious why you need this type of API. Are you committing one message at a time and that (understandably) slows you down?

accelerated:
PS: based on testing with `get_offsets_position()` it results that the offset returned by the rdkafka API is the _same_ as the last message that was returned to the application, which is misleading as the documentation for `rd_kafka_position()` mentions it's the last consumed message + 1.

src/consumer.cpp
```diff
@@ -200,6 +200,24 @@ Consumer::get_offsets_position(const TopicPartitionList& topic_partitions) const
     return convert(topic_list_handle);
 }
 
+void Consumer::store_offsets() const
```
mfontanini: I'm not sure about this overload. When would you use it? Normally you either want to "commit" (in memory in this case) a message or a topic/partition/offset list. Is this meant to be used by just consuming messages, not doing any commit/offset store and then just call this and the offsets get magically committed in memory? If so, I'd probably change this to something like `store_seen_offsets` or `store_consumed_offsets` so it's more in line with what it does. I feel like `store_offsets` without any parameter is not very descriptive.

Also, minor comment: please put braces on the same line as the method definition.

accelerated: This was more as a helper function similar to what the `commit()` does, which uses the assigned partitions under the hood. Since `commit` has 3 overloads I though, for completeness sake, to also have 3 overloads there. I will change it to `store_consumed_offsets`.  
None of the `store_offset` overloads work by themselves, you need to have the `enable.auto.commit=true` so that it actually commits them to the broker. I can also add that in the comment. 

>Also, minor comment: please put braces on the same line as the method definition.

yes, i keep making this mistake all the time :)

accelerated: > I'm not sure about this overload. When would you use it?

Say you consume a batch of messages. Then simply call `store_consumed_offsets()` periodically or after the batch is done.

mfontanini: Fair enough. Looks like [this is failing](https://travis-ci.org/mfontanini/cppkafka/builds/489666899?utm_source=github_status&utm_medium=notification) on rdkafka 0.9.4, so we may need to use some preprocessor guard before using `rd_kafka_offsets_store`. Looks like it was introduced [in v0.11.0[(https://github.com/edenhill/librdkafka/commit/715c49bfbf9f3fe41934989cf92408966949954e)

accelerated: Indeed I hadn't noticed the rdkafka version was different for clang and gcc.

accelerated:
The round robin test is failing intermittently. Tried a few different options but nothing works consistently. Unless I comment the whole test out or at least not check for strict ordering, it won't work. Also running the consumer a 2nd time without the round robin adapter in rdkafka 0.9.4 fails. Messages are not consumed at all. Not sure if it's a timing issue or what?

mfontanini:
Yeah, I've noticed that. We should probably disable the test because it just adds noise and basically we have to always go to travis to see why they're failing and see if failures are related to the PR.

mfontanini:
Thanks sir!

PR #167, opened 2019-02-10
Fix PVS-Studio

pavel-pimenov:
Report: http://www.fly-server.ru/pvs-studio/cppkafka/
V591 	Non-void function should return a value.
V519 	The 'new_item' variable is assigned values twice successively.
V730 	Not all members of a class are initialized inside the constructor.

### DIFF
--- a/include/cppkafka/utils/consumer_dispatcher.h
+++ b/include/cppkafka/utils/consumer_dispatcher.h
@@ -315,7 +315,7 @@ using ConsumerDispatcher = BasicConsumerDispatcher<Consumer>;
 
 template <typename ConsumerType>
 BasicConsumerDispatcher<ConsumerType>::BasicConsumerDispatcher(ConsumerType& consumer)
-: consumer_(consumer),running_(false) {
+: consumer_(consumer) {
 
 }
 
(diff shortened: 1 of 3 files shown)


include/cppkafka/utils/consumer_dispatcher.h
```diff
@@ -315,7 +315,7 @@ using ConsumerDispatcher = BasicConsumerDispatcher<Consumer>;
 
 template <typename ConsumerType>
 BasicConsumerDispatcher<ConsumerType>::BasicConsumerDispatcher(ConsumerType& consumer)
-: consumer_(consumer) {
+: consumer_(consumer),running_(false) {
```
mfontanini: This one is not an issue, everything that checks the value of `running_` is called by `run()` (or one of its callees) which sets it to true.

src/topic_partition_list.cpp
```diff
@@ -45,8 +45,7 @@ TopicPartitionsListPtr convert(const TopicPartitionList& topic_partitions) {
     TopicPartitionsListPtr handle(rd_kafka_topic_partition_list_new(topic_partitions.size()),
                                   &rd_kafka_topic_partition_list_destroy);
     for (const auto& item : topic_partitions) {
-        rd_kafka_topic_partition_t* new_item = nullptr;
-        new_item = rd_kafka_topic_partition_list_add(handle.get(),
+        rd_kafka_topic_partition_t* new_item = rd_kafka_topic_partition_list_add(handle.get(),
```
mfontanini: I'm okay with making this change but only if the formatting looks right. You can make this one look like this, which would look better:

```c++
a* b = function_name(
    param1,
    param2,
    ...
);
```

mfontanini:
Thanks!

PR #171, opened 2019-02-21
Added wait_for_acks(0) when calling async_flush

accelerated:
This is more of an improvement, I'm actually not sure 100% it's worth doing. The rationale for the change is as follows:
* When calling `flush()`, `wait_for_acks()` which essentially calls `Producer::flush()`, is **always** called. This makes sense since we must wait for responses before proceeding. 
* When calling `async_flush()`, `wait_for_acks()` is **never** called, which means that the producer never gets flushed. It's a bit counter-intuitive that one method calls flush but not the other. 
* Inside the producer poll loop, the application must either call `flush()` or the combo `async_flush() + wait_for_acks(0)` if it wants to not block which is easy to miss for someone unfamiliar with this class, and hard to debug.

The solution is to call `wait_for_acks()` in both cases, but not block in the async version of `flush`.

### DIFF
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -221,7 +221,8 @@ public:
     /**
      * \brief Flushes all buffered messages and returns immediately.
      *
-     * Similar to flush, it will send all messages but will not wait for acks to complete.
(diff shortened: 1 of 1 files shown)


mfontanini:
Thanks!

PR #172, opened 2019-03-01
Fix FreeBSD builds

snar:
sys/endian.h already defines all the beNNtoh macros since FreeBSD 5.0,
attempts to redeclare these leads to build failure


mfontanini:
Thanks!

PR #174, opened 2019-03-02
Set CMAKE_CXX_FLAGS only when not set

accelerated:
Allow user overrides  (such as `CXXFLAGS` environtment variable in linux) to completely determine the value of `CMAKE_CXX_FLAGS`. Currently it's impossible to build with `-std-c++14` for example.


mfontanini:
Thanks!

PR #190, opened 2019-04-26
Support for generating pkg-config file

accelerated:
Generates pkg-config file from template `cppkafka.pc.in` and outputs it into `./package/cppkafka.pc`.


mfontanini:
I need to catch up on a bunch of issues and PRs, I'll try to do that tonight. Sorry for the delay on this one!

accelerated:
Thanks!

PR #199, opened 2019-05-20
Removed dependency from rdkafka since it has its own pkg_config file.…

accelerated:
**Description**
* Removed dependency from _rdkafka_ inside the `cppkafka.pc` since _rdkafka_ has its own pkg_config file and it would cause double linking which is unnecessary.
* Also added bitness detection for path installation and library search for the `cppkafka.pc` file.
* Moved dependent libs from `Requires` to `Requires.private` and added comma separation between libs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,16 @@ option(CPPKAFKA_BOOST_STATIC_LIBS "Link with Boost static libraries." ON)
 option(CPPKAFKA_BOOST_USE_MULTITHREADED "Use Boost multithreaded libraries." ON)
 option(CPPKAFKA_RDKAFKA_STATIC_LIB "Link with Rdkafka static library." OFF)
 
+# Determine if this is a 32 or 64 bit build
+string(FIND ${CMAKE_CXX_FLAGS} "-m64" BITNESS)
+
+# Properly set the output directory
+if (${BITNESS} EQUAL -1)
+    set(LIBDIR "lib")
+else()
+    set(LIBDIR "lib64")
+endif()
+
 # Disable output from find_package macro
 if (NOT CPPKAFKA_CMAKE_VERBOSE)
     set(FIND_PACKAGE_QUIET QUIET)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -38,6 +38,16 @@ option(CPPKAFKA_BOOST_STATIC_LIBS "Link with Boost static libraries." ON)
 option(CPPKAFKA_BOOST_USE_MULTITHREADED "Use Boost multithreaded libraries." ON)
 option(CPPKAFKA_RDKAFKA_STATIC_LIB "Link with Rdkafka static library." OFF)
 
+# Determine if this is a 32 or 64 bit build
+string(FIND ${CMAKE_CXX_FLAGS} "-m64" BITNESS)
```
mfontanini: Is there any guarantees there's actually going to be a `-m64` in CXX_FLAGS? I've used something like [this](https://dilawarnotes.wordpress.com/2016/01/13/detect-architecture-of-your-system-using-cmake/) in the past to detect 32/64 bit architectures.

accelerated: Yes I think your proposal is more platform independent. How about something like this which is simpler? from [the 2nd voted answer](https://stackoverflow.com/questions/4029587/compiling-32-bit-vs-64-bit-project-using-cmake)
```
math(EXPR BITS "8*${CMAKE_SIZEOF_VOID_P}")
set(BOOST_LIBRARY "/boost/win${BITS}/lib")
```

mfontanini: Yeah, that would also work

accelerated:
Done. Btw, for the versioning, perhaps you can bump to 1.0.0 since the sonames are not really matching.

mfontanini:
Thanks! What do you mean the sonames aren't matching?

I have a [separate branch](https://github.com/mfontanini/cppkafka/tree/admin-api) on the admin API stuff but I honestly don't have the time (or the willingness) to finish it now. It's actually mostly done I think. I was just missing something to encapsulate the message responses. Maybe I'll push myself to finish it some day soon.

accelerated:
> Thanks! What do you mean the sonames aren't matching?

Sorry i meant the .so versions. Since your version was already tagged in github at a higher number (0.3.x) but your cmake was still building with 0.2

accelerated:
I might give it a go to complete the admin API since we will most likely need it in the near future.

mfontanini:
I already went back and fixed the sonames in 0.3.0 and 0.3.1, that was very broken. 

> I might give it a go to complete the admin API since we will most likely need it in the near future.

Sounds good! Please let me know (comment on #119) when you do so we don't overlap in case I actually start working on it.

accelerated:
sure

PR #200, opened 2019-05-26
Use GNUInstallDirs() over setting up LIBDIR manually

azat:
Setting LIBDIR manually does not allow overwriting it, and for example
archlinux has /lib64 as a symlink to /lib and hence it's package manager
- pacman - will report an error during installation [1]:

    error: failed to commit transaction (conflicting files)
    cppkafka-git: /usr/lib64 exists in filesystem (owned by filesystem)

  [1]: https://aur.archlinux.org/packages/cppkafka-git/#comment-694919

While with GNUInstallDirs() you can overwrite it like this:
  cmake -DCMAKE_INSTALL_LIBDIR=lib ..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,7 @@
-cmake_minimum_required(VERSION 2.8.1)
+cmake_minimum_required(VERSION 2.8.5)
 project(cppkafka)
+
+include(GNUInstallDirs)
 
 # Set the version number.
 set(CPPKAFKA_VERSION_MAJOR 0)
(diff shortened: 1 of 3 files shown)


azat:
Thoughts?

mfontanini:
@accelerated does this work for you? I'm not using pkgconfig and don't normally use it.

accelerated:
I would have to test it and make sure it won't break anything else. It's really important that 64 bit binaries don't end up under `lib`. 
Otherwise I wouldn't be opposed to it.

accelerated:
I just got around to testing with this...works well! Definitely makes things simpler as you don't have to manually generate the bitness. I would say it's good to merge.

accelerated:
PS: I have a PR which incorporates this as well as more CMAKE improvements and modernizations...so no need to take this PR in for now.

accelerated:
This one should be closed now...

azat:
Indeed, everything works now

PR #210, opened 2019-07-05
Remove boost dependency from pkg-config template file

ych:
Boost does not provide pkg-config file, so if execute
 'pkg-config --exists cppkafka' command with boost dependency,
 user always gets non-zero return. And PKG_SEARCH_MODULE in cmake
 use the command to check the status of cppkafka.

The boost dependency should be removed for general usage can be works.

### DIFF
--- a/cmake/cppkafka.pc.in
+++ b/cmake/cppkafka.pc.in
@@ -11,4 +11,4 @@ Version: @CPPKAFKA_VERSION@
 Requires:
 Requires.private: rdkafka >= 0.9.4
 Libs: -L${libdir} -L${sharedlibdir} -lcppkafka
-Cflags: -I${includedir} -I${includedir}/cppkafka
+Cflags: -I${includedir} -I${includedir}/cppkafka -I@Boost_INCLUDE_DIRS@
--- a/cmake/cppkafka.pc.in
+++ b/cmake/cppkafka.pc.in
@@ -9,6 +9,6 @@ Url: https://github.com/mfontanini/cppkafka
 Description: C++ wrapper library on top of RdKafka
 Version: @CPPKAFKA_VERSION@
 Requires:
-Requires.private: rdkafka >= 0.9.4, boost
+Requires.private: rdkafka >= 0.9.4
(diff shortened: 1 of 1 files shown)


accelerated:
@ych, do you have to add boost manually on the link line? Maybe there's a flag in pkg-config to skip unknown .pc files instead of throwing an error? Our boost installs have .pc files for instance, but I think it's more of a company artifact than something standard.

ych:
@accelerated  If cppkafka really need Boost but Boost is not supported in pkg-config, it should put to Cflags field. (please refer https://people.freedesktop.org/~dbn/pkg-config-guide.html, "Cflags: The compiler flags specific to this package and any required libraries that don't support pkg-config.")

But I try to find out where cppkafka using Boost, and I find out that boost only is used in example for program options. And pkg-config is a way to describe how to use the library, it does not need to include the library which only example program used.

accelerated:
@ych, in that case we have to bring back `-I@Boost_INCLUDE_DIRS@` in CFlags because CppKafka uses `boost::optional` which is header-only file. `boost::program_options` are used in tests only. This is what I originally had but i was unaware that open source boost distros don't provide a .pc file. So my testing was not failing because our company generates these .pc files.

ych:
@accelerated ok, I only use ldd command to check executable files and library files, so I ignore the `boost::optional` part.

I add `@Boost_INCLUDE_DIRS@` to pkg-config template file, it will be converted by `configure_file` command in CMakeLists.txt.

accelerated:
@ych, can you please fix the conflict so this can be merged?

ych:
I am sorry to have noticed the comment.

The conflict should be fixed now.

mfontanini:
@accelerated are you good with this one? I don't use pkg-config so I'm not sure what are the implications of this.

accelerated:
Yes I recently heard from another source that pkg_config if broken if used with cmake because of boost, so I say it should be removed.

mfontanini:
Perfect. Thanks @ych and @accelerated!

PR #215, opened 2019-07-26
Added method to set the partition

accelerated:


mfontanini:
I'm not really a fan of having setters for everything. In which context do you not know the partition during construction? The (topic, partition) is basically an identifier for the object, that probably shouldn't be changing.

accelerated:
In this scenario, we have a consumer cluster reading a topic, the configuration is the same across the cluster. However each instance gets a static partition assignment either in a separate file or from db. Therefore we create the global config first, then use setters for the respective partitions and offests. So it's nice to have, otherwise you need to delay the creation of these objects.
Secondly, you have a constructor that only takes a topic. How are you supposed to set the partition in that case?

mfontanini:
Fair enough! I guess normally I'd just keep the topic name and build the `TopicPartition`s when needed but I understand if you may want to avoid the `string` memory allocations after whatever setup phase you have.

accelerated:
Thanks!

PR #223, opened 2019-09-17
update CMakeList.txt

chujiangke:
update examples CMakeList.txt


accelerated:
Hi @chujiangke what is the reason for this change? By adding executables directly they will appear as targets in your makefile which is what @mfontanini doesn't want. Also there is no linking provided for each target and the dependency should be on `RdKafka::rdkafka` imported library not `${RDKAFKA_LIBRARY}` directly.

chujiangke:
because Ubuntu can't  build the example， so I change it。

accelerated:
Could you post the error you're getting? This PR is definitely not the way to go.

chujiangke:
cmakelist write like below
‘’‘create_example(producer)
create_example(buffered_producer)
create_example(consumer)
create_example(consumer_dispatcher)
create_example(metadata)
create_example(consumers_information)’

do cmake .
make 

and example  make nothing , so i change it , it works well.

accelerated:
You need to explicitly call `make examples` and it will work. By default examples are not part of `make all`

chujiangke:
oh thinks 。

PR #231, opened 2020-01-06
Added RdKafkaException base class

accelerated:
* Added `RdKafkaException` base class for exceptions which take an `Error` as parameter. This allows for easier catching and processing of kafka-specific errors.
* In cases where the error message was sinked into the Exception (`ParseException` and `ActionTerminatedException`), the signature is _pass-by-copy_ now to be consistent with the base `Exception` class. This should not break ABI since the constructor is only called internally.

### DIFF
--- a/include/cppkafka/exceptions.h
+++ b/include/cppkafka/exceptions.h
@@ -51,6 +51,19 @@ private:
 };
 
 /**
+ * Base class for all rdkafka exceptions
+ */
+class CPPKAFKA_API RdKafkaException : public Exception {
+public:
+    RdKafkaException(Error error);
+    RdKafkaException(Error error, std::string message);
+    
+    Error get_error() const noexcept;
+private:
+    Error error_;
+};
+
+/**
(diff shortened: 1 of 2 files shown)


accelerated:
Hi @mfontanini, can you please take a look at this? ~thanks

mfontanini:
While I'm okay with these changes and this should probably have been done this way from the start, this change breaks the ABI and API. This:

* Assumes nobody is throwing these exceptions manually (e.g. while testing code).
* (potentially) changes the layout of these exceptions. I think in practice they'll end up having the same layout but it's a risky change.

Having said that, I don't think it's worth bumping the major version for such a simple change.

accelerated:
Agreed with you. I think it's a worthwhile change with small breakage risk.

accelerated:
@mfontanini are you ok going forward with this? Can we merge it? Otherwise I'll close it.

accelerated:
@mfontanini please check my last comment.

PR #235, opened 2020-02-03
Fix tracker promise from throwing when set multiple times

accelerated:
Two issues:
* `std::deque::size()` should be guarded as it's not thread safe. The size() is computed via pointer diff (finish-start) of the queue itself and it returns garbage in a multi-threaded access scenario. 
* The tracker promise may be set multiple times if there are retries. If the produce is asynchronous, the future of this promise is never checked (and as such the promise is never re-created), which means the **same** promise can be set multiple times, throwing a `promise_already_satisfied` exception. This in itself is not an error, however it prevents the `pending_acks` from being properly decremented and the flush eventually blocks forever as the `pending_acks` mismatch. 

@mfontanini please prioritize this if you can. Much appreciated!!

### DIFF
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -771,7 +771,7 @@ void BufferedProducer<BufferType, Allocator>::clear() {
 
 template <typename BufferType, typename Allocator>
 size_t BufferedProducer<BufferType, Allocator>::get_buffer_size() const {
-    int size = 0;
+    size_t size = 0;
     {
         std::lock_guard<std::mutex> lock(mutex_);
         size += messages_.size();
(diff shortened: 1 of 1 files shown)


include/cppkafka/utils/buffered_producer.h
```diff
@@ -771,7 +771,16 @@ void BufferedProducer<BufferType, Allocator>::clear() {
 
 template <typename BufferType, typename Allocator>
 size_t BufferedProducer<BufferType, Allocator>::get_buffer_size() const {
-    return messages_.size() + retry_messages_.size();
+    int size = 0;
```
mfontanini: `size_t` here

accelerated:
Yes it has gotten a bit out of hand with too many features instead of keeping a strict delineation of functionality and split it into smaller classes. The class API however is quite clean so that's always good. I may revisit the inner implementation the future or perhaps bring back the original design under some other name and keep this one as-is (if you want). Not too many use it so hopefully it's not causing instability elsewhere.
I think however that admin API would have higher priority as future work.

mfontanini:
Yeah, the admin API should be the next thing although I probably won't for on that in the near future.

PR #238, opened 2020-02-09
Added member functions for static consumers

accelerated:
Added `assign` and `revoke` virtual functions for static consumers in `PollStrategyBase` class. Currently there is no way to assign or remove partitions for statically assigned consumers.

### DIFF
--- a/include/cppkafka/utils/poll_strategy_base.h
+++ b/include/cppkafka/utils/poll_strategy_base.h
@@ -84,6 +84,29 @@ public:
      */
     Consumer& get_consumer() final;
     
+    /**
+     * \brief Creates partitions queues associated with the supplied partitions.
+     *
+     * This method contains a default implementation. It adds all the new queues belonging
+     * to the provided partition list and calls reset_state().
+     * To be used with static consumers.
+     *
+     * \param partitions Assigned topic partitions.
+     */
+    virtual void assign(TopicPartitionList& partitions);
+    
+    /**
+     * \brief Removes partitions queues associated with the supplied partitions.
+     *
+     * This method contains a default implementation. It removes all the queues
+     * belonging to the provided partition list and calls reset_state().
+     * To be used with static consumers.
+     *
+     * \param partitions Revoked topic partitions. If the partition list is empty
+     *                   all partitions will be revoked.
+     */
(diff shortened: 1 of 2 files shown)


src/utils/poll_strategy_base.cpp
```diff
@@ -89,29 +89,43 @@ void PollStrategyBase::reset_state() {
...
+
+void PollStrategyBase::revoke(const TopicPartitionList& partitions) {
+    if (partitions.empty()) {
+        //revoke everything
+        partition_queues_.clear();
+    }
+    else {
+        for (const auto &partition : partitions) {
+            // get the queue associated with this partition
+            auto toppar_it = partition_queues_.find(partition);
```
mfontanini: I think you can directaly call `partition_queues_.erase(partition)` here

include/cppkafka/utils/poll_strategy_base.h
```diff
@@ -84,6 +84,29 @@ class CPPKAFKA_API PollStrategyBase : public PollInterface {
...
+    
+    /**
+     * \brief Removes partitions queues associated with the supplied partitions.
+     *
+     * This method contains a default implementation. It removes all the queues
+     * belonging to the provided partition list and calls reset_state().
+     * To be used with static consumers.
+     *
+     * \param partitions Revoked topic partitions. If the partition list is empty
+     *                   all partitions will be revoked.
```
mfontanini: What's the reason for this behavior of removing all if the list is empty? That feels a bit counter intuitive. You might as well add a `revoke()` that takes nothing which makes the semantics a bit more clear from the user's perspective

src/utils/poll_strategy_base.cpp
accelerated: True, i'm not even sure where the find comes from originally. Must have been a remnant from some other action which was applied before erasing.

accelerated:
@mfontanini can you please finalize this when you get a chance? Thanks!

accelerated:
Hi @mfontanini, any updates on these PRs?

mfontanini:
Sorry, I haven't been paying much attention to github emails these days, I didn't remember there were things pending.

accelerated:
@mfontanini Thanks. Please check the other PRs as well.

PR #247, opened 2020-05-21
Add support for rd_kafka_destroy_flags.

filimonov:
That allows to set flags for rd_kafka_destroy_flags in the following manner:
```cpp
     consumer.set_destroy_flags(RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE);
```

It allows us to prevents potential hang during termination if consumer.close() is failing because of queued callback calls.

Refs:
- https://github.com/edenhill/librdkafka/blob/7269f0c6a225cd42dc7a44c5e081c464986d20ff/tests/0084-destroy_flags.c#L69
- https://github.com/edenhill/librdkafka/issues/2077

More:
- Maybe it can also be used as a default (because Consumer destructor call consumer_close) 
- or that flag maybe enabled automatically if an exception happens during consumer_close call in consumer destructor.
- Most probably that test can be enabled: https://github.com/mfontanini/cppkafka/blob/006642cdb2a871ef3aad517e9148607b859604e0/tests/consumer_test.cpp#L245-L260

Many thanks to @azat for help.


include/cppkafka/kafka_handle_base.h
```diff
@@ -340,6 +340,30 @@ class CPPKAFKA_API KafkaHandleBase {
      */
     int get_out_queue_length() const;
 
+#if RD_KAFKA_VERSION >= RD_KAFKA_DESTROY_FLAGS_SUPPORT_VERSION
+    /**
+     * \brief Sets flags for rd_kafka_destroy_flags()
+     *
+     * 0 (default) - calls consumer_close() during handle destruction
```
accelerated, line 347: I would only keep the `brief` message. The rest is too specific for one particular flag and there are other flags also.

include/cppkafka/kafka_handle_base.h
```diff
@@ -357,7 +381,20 @@ class CPPKAFKA_API KafkaHandleBase {
 private:
     static const std::chrono::milliseconds DEFAULT_TIMEOUT;
 
-    using HandlePtr = std::unique_ptr<rd_kafka_t, decltype(&rd_kafka_destroy)>;
+    // It seems that destroy_flags maybe RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE by default.
+    //
+    // All the consumer objects created by cppkafka call rd_kafka_consumer_close during destruction.
+    // So we don't want it to be called once again during handle destruction.
+    int destroy_flags_ = 0;
```
accelerated, line 388: `destroy_flags_` should go at the bottom of the class after `handle_`. Also use brace-list-initializer instead of `=`.  I would also remove all the comments pertaining to the flags, since they are particular to one flag only and there are other flags.

include/cppkafka/kafka_handle_base.h
```diff
@@ -357,7 +381,20 @@ class CPPKAFKA_API KafkaHandleBase {
...
+
+    struct handle_deleter {
```
accelerated, line 390: Use camel case for class names i.e. `HandleDeleter`. 
Make constructor `explicit`. 
I would change to `const KafkaHandleBase& handle_base_` because having this as nullptr is not valid.

src/kafka_handle_base.cpp
```diff
@@ -48,7 +48,7 @@ namespace cppkafka {
 const milliseconds KafkaHandleBase::DEFAULT_TIMEOUT{1000};
 
 KafkaHandleBase::KafkaHandleBase(Configuration config) 
-: timeout_ms_(DEFAULT_TIMEOUT), config_(move(config)), handle_(nullptr, nullptr) {
+: timeout_ms_(DEFAULT_TIMEOUT), config_(move(config)), handle_(nullptr, handle_deleter(this)) {
```
accelerated, line 51: keep `nullptr` here. Otherwise `rd_kafka_delete_flags` will be called with null handle and application will crash.

src/kafka_handle_base.cpp
```diff
@@ -213,7 +213,7 @@ void KafkaHandleBase::yield() const {
 }
 
 void KafkaHandleBase::set_handle(rd_kafka_t* handle) {
-    handle_ = HandlePtr(handle, &rd_kafka_destroy);
+    handle_ = HandlePtr(handle, handle_deleter(this));
```
accelerated, line 216: Pass `*this`

filimonov:
BTW: it didn't help with the original problem :\ Still need draining after unsubscribe: https://github.com/edenhill/librdkafka/issues/2898

accelerated:
Why do you need to connect 2 consumers inside the same app to the same topic?

filimonov:
3 reasons:
1) actually i was just testing the behavior after rebalance  and if was the simplest option to trigger rebalance
1) it's out of my control, users of ClickHouse can create several 'tables' = 'kafka consumers'. I can't rectict that.
2) parallell consumtion is done by running several consumers instead of several queues connected to same consumer (legacy code, should be rewritten, bit tricky).

src/kafka_handle_base.cpp
filimonov: That doesn't work:
```
error: no matching constructor for initialization of 'cppkafka::KafkaHandleBase::HandlePtr' (aka 'unique_ptr<rd_kafka_s, cppkafka::KafkaHandleBase::HandleDeleter>')
[build] : timeout_ms_(DEFAULT_TIMEOUT), config_(move(config)), handle_(nullptr, nullptr), destroy_flags_(0) {
```

https://stackoverflow.com/questions/50201967/why-unique-ptr-with-custom-deleter-wont-work-for-nullptr-while-shared-ptr-does

I think it's quite safe to use this here.

accelerated: Ah yes, correct, it's changed to a functor instead of a func ptr. Ok but put a check inside operator() for the handle being null.

accelerated:
I would recommend not having multiple consumers in the same app if you can. It's very counterintuitive and you may run into other issues down the line.

src/kafka_handle_base.cpp
azat: >Otherwise rd_kafka_delete_flags will be called with null handle and application will crash.

unique_ptr will not call deleter if there is no object, or I'm missing something?

accelerated: Seems to say that windows does not behave properly.

filimonov:
> I would recommend not having multiple consumers in the same app if you can.

I can't :)

src/kafka_handle_base.cpp
filimonov: That change seems a bit problematic, and I need to switch to other tasks. 
I'm totally ok if you will improve that.

src/kafka_handle_base.cpp
filimonov: discussed above.

src/kafka_handle_base.cpp
accelerated: > > Otherwise rd_kafka_delete_flags will be called with null handle and application will crash.
> 
> unique_ptr will not call deleter if there is no object, or I'm missing something?
The link posted seems to imply that on Windows, the unique_ptr is not implemented properly.

accelerated:
@filimonov Also, since `HandleDeleter` cannot be set with a null `KafkaHandleBase` I would recommend capturing it by reference. It makes intent clearer in the class design.

filimonov:
I've tried, but it seem like it will require deeper look to do that
```
[build] /home/mfilimonov/workspace/ClickHouse/contrib/libcxx/include/memory:2538:21: error: object of type 'cppkafka::KafkaHandleBase::HandleDeleter' cannot be assigned because its copy assignment operator is implicitly deleted
[build]     __ptr_.second() = _VSTD::forward<deleter_type>(__u.get_deleter());
[build]                     ^
[build] /home/mfilimonov/workspace/ClickHouse/contrib/cppkafka/src/kafka_handle_base.cpp:216:13: note: in instantiation of member function 'std::__1::unique_ptr<rd_kafka_s, cppkafka::KafkaHandleBase::HandleDeleter>::operator=' requested here
[build]     handle_ = HandlePtr(handle, HandleDeleter(*this));
```
If you can improve that - please do.

src/kafka_handle_base.cpp
filimonov: You're looking on the question in the linked SO thread, check the answer.

accelerated:
Yeah I just see there's a requirement for the deleter to be default constructible so yes, you can't use any references unfortunately.

mfontanini:
Thanks for the PR!

PR #248, opened 2020-05-21
Try to fix travis

filimonov:
Old URL returns 404, picked new from official webpage: https://kafka.apache.org/downloads

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ compiler:
 env:
   - RDKAFKA_VERSION=v0.9.4
   - RDKAFKA_VERSION=v0.11.5
-  - RDKAFKA_VERSION=v1.1.0
+  - RDKAFKA_VERSION=v1.0.0
 
 os:
   - linux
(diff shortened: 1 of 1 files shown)


filimonov:
@accelerated  do you know smth about why the travis tests fail with the newest librdkafka?

accelerated:
Looks to me like a race condition in the compacted processor and it gets stuck. Unfortunately I dont have access to Travis to restart the builds since i'm not the repo owner

mfontanini:
Restarted the build but the same thing is probably going to happen.

mfontanini:
Yeah, this is hanging on the new kafka version. Can you remove the new broker version in the travis matrix? We should figure out what's up with that test but that can wait.

filimonov:
Let's try a bit older version first...

filimonov:
v1.3.0 is failing too...

filimonov:
So the problem started with https://github.com/edenhill/librdkafka/releases/tag/v1.0.0

stevenylai:
The test is stuck when the consumer is trying to wait for eof for all partitions. I briefly tested with the new library, we need to first set: "enable.partition.eof": true, and second, rdkafka only returns eof to those are assigned to the consumer. I will try to open a PR and see if it fixes this

mfontanini:
Thanks for the fix!

Issue #249, opened 2020-05-24
compacted_topic_processor_test hangs with librdkafka > 1.0.0


filimonov:
Running test "consumption" @ /home/travis/build/mfontanini/cppkafka/tests/compacted_topic_processor_test.cpp:47

See Travis test runs in https://github.com/mfontanini/cppkafka/pull/248/

PR #251, opened 2020-06-15
Fix tests

stevenylai:
This should fix https://github.com/mfontanini/cppkafka/issues/249

I also added NODE_ID Macro so that I can use my existing cluster to run those tests

It should be based on the other travis pull request but I don't think I can merge from one fork to another.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -5,6 +5,10 @@ if (NOT KAFKA_TEST_INSTANCE)
 set(KAFKA_TEST_INSTANCE kafka-vm:9092
     CACHE STRING "The kafka instance to which to connect to run tests")
 endif()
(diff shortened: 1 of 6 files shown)


jlcordeiro:
I had the same problems whilst running the tests and these changes did indeed fix them.
PR build seems to be failing but at first glance it was a temporary issue (?)

jlcordeiro:
@stevenylai did a few extra tests and everything build fine to me.
I'd suggest maybe closing / reopening PR to see if we get all checks passing before suggesting a merge onto master.

stevenylai:
@jlcordeiro Yes, I wanted to check this but always forgot afterwards. Let me see if I can update the branch to trigger a re-build

stevenylai:
I see the current master is already passing travis build so I guess there is nothing to be fixed then. Closing.

PR #257, opened 2020-07-19
Bring base class  resume into scope

accelerated:
Bring base class `resume()` member function into the scope of `Consumer`. This was missing since `pause` is already into scope.


accelerated:
@mfontanini if you can merge this it would be great! Currently access to base class overload is unavailable.

mfontanini:
:pray:

PR #261, opened 2020-08-14
Ability to block producer in case the underlying queue is full.

hanickadot:
Make the flag accessible to cppkafka API.


mfontanini:
Thanks!

PR #263, opened 2020-09-01
Call flush termination callbacks from sync_produce

accelerated:
** Description **
Call flush termination callbacks from sync_produce. This allows `sync_produce()` to exhibit the same behavior as `async_produce()` when the `rd_kafka` underlying layer throws an error upon production. Flush callbacks, if registered, will now be called in both cases. 
If the application passed some opaque correlation data via the `user_data` pointer and expected the delivery callback to be triggered on success or failure, in the failure case, this was not guaranteed to happen depending on the error type. For instance producing to a bad topic name, would cause all flushes to throw and no callbacks were ever invoked.

_Note that sync_produce as well as async_produce can also be called during a flush, not only directly by the producer._

### DIFF
--- a/include/cppkafka/utils/buffered_producer.h
+++ b/include/cppkafka/utils/buffered_producer.h
@@ -691,6 +691,7 @@ private:
     void do_add_message(BuilderType&& builder, QueueKind queue_kind, FlushAction flush_action);
     template <typename BuilderType>
     void produce_message(BuilderType&& builder);
+    bool sync_produce(const MessageBuilder& builder, std::chrono::milliseconds timeout, bool throw_on_error);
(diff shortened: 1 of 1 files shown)


mfontanini:
I can merge this now if it's ready, I can't follow what this class is doing anymore.

accelerated:
Sure, thanks! BTW I think the mirror for the rdkafka tarball is broken. There's a fix for this, you should consider merging as it's hard to tell w/o the CI tests.

mfontanini:
Sorry, I thought there were still other problems on that one. Merged it, can you rebase when you have a chance?

accelerated:
@mfontanini it's good to go!

mfontanini:
Thanks sir!

PR #265, opened 2020-09-04
Fix CMAKE warning

accelerated:
Fixes the following warning:
```
The package name passed to `find_package_handle_standard_args` (RDKAFKA)
  does not match the name of the calling package (RdKafka).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
```


mfontanini:
:ok_hand:

PR #272, opened 2021-03-04
Add fix for in-source build of cppkafka.h.

psigen:
This fixes an issue where `cppkafka.h` is being generated directly within the source directory, which causes issues on sandboxed build environments where the source directory is mounted readonly.

This PR changes the `configure_file()` directive to point to the cmake binary directory, and uses the install directives to move the generated file into the build output.

I also added quoting around constructed file paths that could ostensibly include spaces, to prevent any list parsing issues.


mfontanini:
Sorry it took me some time to get to this. I think there was _some_ reason why this was done this way initially, knowing it would pollute the source directory, but I don't really remember what it was (maybe allowing doing `#include`s directly into the source dir? I don't know).

Anyway, this is a better solution for sure. Thanks!

PR #291, opened 2022-03-20
cmake: Use CMAKE_INSTALL_LIBDIR

kraj:
Do not hardcode library path, this helps it install in right location
independent of platforms, e.g. ppc64 uses lib64 and not lib

Signed-off-by: Khem Raj <raj.khem@gmail.com>


mfontanini:
Thanks!

Issue #206, opened 2019-07-02
C2491 “cppkafka::BufferedProducer<BufferType,Allocator>::BufferedProducer” definition of dllimport function not allowed


latermarch:
![image](https://user-images.githubusercontent.com/9695216/60475774-9387c500-9cab-11e9-9779-57c171166ef5.png)
win10 vs2015
I git clone the latest project from the master

latermarch:
anyone helps

accelerated:
Try defining `cppkafka_EXPORTS`

latermarch:
@accelerated 
 thanks ,I tried, and delete the function in  header buffered_producer.h
![image](https://user-images.githubusercontent.com/9695216/60559867-7670f600-9d80-11e9-8363-3623ea6c3fe2.png)

steinio:
Hi. I just got this very issue yesterday. Do I define `cppkafka_EXPORTS` in macros.h, before the conditionals that uses it, and do I define this before building the library as well? Also, his solution of commenting out the method that causes the error, are there any cleaner ways around this problem?

accelerated:
You would need to define it as a CMAKE build parameter such as `-DCMAKE_CXX_FLAGS="<default ompile options> -Dcppkafka_EXPORTS"`. If you're on windows, the default suggested options are [here](https://github.com/mfontanini/cppkafka/blob/master/CMakeLists.txt#L13-L22). In visual studio you just put the -Dcppakfka_EXPORTS I would imagine in the same place as you would in a normal project built by VS compiler (`Project's Properties -> C/C++ -> Preprocessor definitions`).

I see the error explained [here](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2491?view=vs-2019) but the source code in the Buffered Producer seems OK to me. Anyway, try adding the definition above.

NOTE: IF you are trying to build the library, then `cppkafka_EXPORTS` has to be defined. If you are linking you application with the `cppkafka.dll`, then you MUST NOT define this since you are now _importing_ the symbols.

steinio:
This works for me btw, the fix you mentioned. I ran into issues when I tried to use the code that @latermarch commented out, bufferedproducer, which gave an "undecklared identifier for queue1" as well as queue2. I fixed this by simply removing the identifier as show below:

`template <typename BufferType, typename Allocator>
[...]

PR #292, opened 2022-04-07
Do not try to export template declaration in shared lib

SpaceIm:
It doesn't make sense to export a template declaration.

closes https://github.com/mfontanini/cppkafka/issues/206


mfontanini:
Thanks!

PR #293, opened 2022-04-07
CMake: set min required C++ standard to C++11

SpaceIm:
do not hardcode -std=c++11, but let CMake set the minimum required C++ standard of cppkafka if consumer does not force CMAKE_CXX_STANDARD


mfontanini:
Thanks!

mfontanini:
Thanks!

PR #307, opened 2023-06-13
[Memory Leak] Only release header handle if producev returned no errors

BrieucNx:
There is a memory leak that happens when the `rd_kafka_producev` function returns an error. This is due to the fact that the function itself takes a pointer to the header struct and takes ownership of it only if no error occurs. If the function returns something else that `RD_KAFKA_RESP_ERR_NO_ERROR` then the ownership of `headers` is not taken.

This patch passes the pointer of the header to the produce function and only release the ownership of the memory if the function returned an non error error code.


mfontanini (approved):
Nice, thanks!

Light7734:
Hey when are you guys releasing this? this is not available on any of the conan center recipes, thanks.

mfontanini:
Just released this https://github.com/mfontanini/cppkafka/releases/tag/v0.4.1. Sorry for the delay!

PR #309, opened 2023-07-13
get_member_id leak fix

ilejn:
According to librdkafka manual, 
the application must free the area returned by rd_kafka_memberid 
with free() or rd_kafka_mem_free()


mfontanini:
Thanks!

PR #318, opened 2025-01-08
Do not link unneeded libraries

georgthegreat:
These libraries are needed by librdkafka, not by cppkafka, hence there is no need to link them.

Removing this fixes NixOS build of the library (NixOs does not provide zlib / openssl dependencies by default).


cmake/FindRdKafka.cmake
```diff
@@ -56,9 +56,9 @@ try_compile(RdKafka_FOUND ${CMAKE_CURRENT_BINARY_DIR}
 if (RdKafka_FOUND)
     add_library(RdKafka::rdkafka ${RDKAFKA_LIBRARY_TYPE} IMPORTED GLOBAL)
     if (UNIX AND NOT APPLE)
-        set(RDKAFKA_DEPENDENCIES pthread rt ssl crypto dl z)
+        set(RDKAFKA_DEPENDENCIES pthread rt dl)
```
fdr400, line 59: This breaks if librdkafka is found as a static library
On the other hand, the trunk versions is also broken for static librdkafka, because it requires more dependencides (at least sasl2, zstd, lz4, curl depending on the librdkafka build flags)

georgthegreat: These are librdkafka dependencies, not cppkafka ones.
cppkafka possess no (??) knowledge regarding librdkafka configuration and hence is unable to guess the necessary libraries to be linked.

fdr400: yes, but it is not possible to link statically if library does not provide its cmake config => you should find all dependencies manually and link target executable to them

on Ubuntu, librdkafka debian package provides only headers and .a/.so files without RdKafkaConfig.cmake (like many other packages), though if find_library returns librdkafka.a, it must also link with all dependencies explicitly

by the way, packages often have config file to pkg-config, which can tell which libraries is requires
proper Find cmake module should link with them through pkg-config --libs --static <lib-name>

but the easiest reliable solution, i believe, is to substitute `find_library(NAMES rdkafka)` with `find_library(NAMES librdkafka.so librdkafka.dylib)` and place `find_package(RdKafka CONFIG) if (RdKafka_FOUND) return()` on the top of the FindRdKafka.cmake (to try to find manually installed package at first)

georgthegreat: `pkg-config --libs rdkafka` is the tool for listing necessary libraries.

Stop saying nonsence, please.
This PR is not about librdkafka.

PR #322, opened 2026-02-26
Add OAuth bearer token refresh callback support

kalavt:
This PR adds support for OAuth bearer token refresh callbacks, enabling authentication mechanisms like AWS MSK IAM, Azure Event Hubs, and custom OAuth implementations.

## Changes:
- Add OAuthBearerTokenRefreshCallback support to Configuration class
- Implement callback proxy for rd_kafka_conf_set_oauthbearer_token_refresh_cb
- Add comprehensive documentation and example
- Use pointer parameter for semantic consistency with librdkafka

## Implementation Details:
The callback uses `const std::string*` instead of `const std::string&` to maintain semantic consistency with librdkafka, allowing callers to distinguish between:
- Config not set (nullptr)
- Config set to empty string ("")

## Testing:
Includes complete example implementation in examples/oauth_example.cpp

### DIFF
--- a/docs/oauth_bearer_callback.md
+++ b/docs/oauth_bearer_callback.md
@@ -1,0 +1,211 @@
+# OAuth Bearer Token Refresh Callback
+
+## Overview
+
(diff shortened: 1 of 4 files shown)


kalavt:
@mfontanini would you pls review the changes, for add oauth refresh callback

